PR #5, opened 2015-01-17
Now selectively loads planners and does not fail when it can't find one.

aaronwalsman:


psigen:
Ok, this looks good. Adding.

PR #6, opened 2015-02-09
Updates TSRs and adding script to publish head TF frames

jeking04:
This branch has updates to the TSRs for the bowl, plate and glass used for the table clearing demo. It also has the head tf publisher. This branched off of feature/table_clearing_demo so once it is pulled in we can delete both branches.


mkoval:
I'm still not a fan of the head publisher script. Unfortunately, the "right" answer, i.e. modifying OWD, is far more trouble than it's worth. So, I'l go ahead and merge this.

PR #7, opened 2015-02-09
Transmission ratio calibration script

mkoval:
This branch includes a script for calibrating HERB's WAM's transmission ratios using an external US Digital X3 three-axis inclinometer. We've now used this to successfully calibrate both arms.

The script still requires a bit of clean-up before I can merge it.


mkoval:
This works. So there's no point in keeping it in a branch.

PR #9, opened 2015-03-13
Set HERB's acceleration limits (not set by URDF).

jeking04:
@siddhss5 Can you double check that these accelerations look sane?


siddhss5:
Almost impossible to tell from this:
http://support.barrett.com/wiki/WAM/SafetyFeaturesPrecautions

It looks like:
0.5 r/s = joint speed warning
2 r/s = joint speed fault
0.5 m/s = elbow speed warning [elbow at 0.55m from J0]
1 m/s = elbow speed fault
4731 magic units = torque warning
5231 magic units = torque fault

So, 2 r/s^2 for joint accelerations should be ok ... ?
Ideally, we should take those above values and make a simple Python file that checks for vel and acc warnings and faults given a trajectory. And make sure we pass any timed trajectory through that before execution.

PR #13, opened 2015-03-27
Feature/pointing tsr

rachelholladay:
Add fuze.py to enable pointing

### DIFF
--- a/src/herbpy/tsr/fuze.py
+++ b/src/herbpy/tsr/fuze.py
@@ -41,7 +41,7 @@ def point_at_obj(robot, bottle, manip=None):
     Bw_0 = numpy.zeros((6, 2))
     Bw_0[3, :] = [-numpy.pi, numpy.pi]
     Bw_0[4, :] = [0, numpy.pi]
-    Bw_0[5, :] = [-numpy.pi, numpy.py]
+    Bw_0[5, :] = [-numpy.pi, numpy.pi]
 
     Bw_1 = numpy.zeros((6, 2))
     Bw_1[2, :] = [-0.75, 1.]
--- a/src/herbpy/tsr/fuze.py
+++ b/src/herbpy/tsr/fuze.py
@@ -25,9 +25,9 @@ def point_at_obj(robot, bottle, manip=None):
     # compute T_we. Tuned once and used forever onward 
     # Insures pointer points at object with right arm
     arm_start =  numpy.array([[ 0.07277, -0.71135,  0.69905,  0.58575],
-       			              [ 0.03360,  0.70226,  0.71111, -0.44275],
-    					      [-0.99678, -0.02825,  0.07501,  0.95686],
-   					          [ 0.     ,  0.     ,  0.     ,  1.     ]])
+       			      [ 0.03360,  0.70226,  0.71111, -0.44275],
+    		              [-0.99678, -0.02825,  0.07501,  0.95686],
+ 			      [ 0.     ,  0.     ,  0.     ,  1.     ]])
 		
     base_obj_trans = numpy.array([[1, 0, 0, 1.0],
                                   [0, 1, 0, 0.0],
--- a/src/herbpy/tsr/fuze.py
+++ b/src/herbpy/tsr/fuze.py
@@ -24,14 +24,14 @@ def point_at_obj(robot, bottle, manip='right'):
     # Ensures pointer points at object with right arm
     # TODO need to add increased documentation once resolved 
     arm_start =  numpy.array([[ 0.07277, -0.71135,  0.69905,  0.58575],
-             			      [ 0.03360,  0.70226,  0.71111, -0.44275],
-    		                  [-0.99678, -0.02825,  0.07501,  0.95686],
- 			                  [ 0.     ,  0.     ,  0.     ,  1.     ]])
+                              [ 0.03360,  0.70226,  0.71111, -0.44275],
+                              [-0.99678, -0.02825,  0.07501,  0.95686],
+                              [ 0.     ,  0.     ,  0.     ,  1.     ]])
 		
     T_base_obj = numpy.array([[1, 0, 0, 1.0],
-                                  [0, 1, 0, 0.0],
-                                  [0, 0, 1, 1.0],
-                                  [0, 0, 0, 1.0]])
+                              [0, 1, 0, 0.0],
+                              [0, 0, 1, 1.0],
+                              [0, 0, 0, 1.0]])
 
     TW_e_0 = numpy.dot(numpy.linalg.inv(T_base_obj), arm_start)
     TW_e_1 = numpy.identity(4)
--- a/src/herbpy/tsr/fuze.py
+++ b/src/herbpy/tsr/fuze.py
@@ -1,39 +1,32 @@
-# TODO fix the imports
 import numpy
 import prpy.tsr 
 
 @prpy.tsr.tsrlibrary.TSRFactory('herb', 'fuze_bottle', 'point')
-def point_at_obj(robot, bottle, manip='right'):
+def point_at_obj(robot, bottle, manip=None):
     '''
     @param robot The robot performing the point
     @param bottle The bottle to point at
     @param manip The manipulator to point with. This must be the right arm 
     '''
-    if manip.GetName() == 'right':
-        with manip.GetRobot():
-            manip.SetActive()
-            manip_idx = manip.GetRobot().GetActiveManipulatorIndex()
-    else:
-        raise Exception("The pointing tsr is only defined for the right manipulator")
+    if manip is None:
+        manip = robot.right_arm
+
+    manip.SetActive()
+    manip_idx = manip.GetRobot().GetActiveManipulatorIndex()
+    
+    if manip.GetName() != 'right':
+        raise prpy.exceptions.PrPyException('Pointing is only defined on the right arm.')
 		
     # compute T_ow
     T0_w_0 = bottle.GetTransform()
     T0_w_1 = numpy.identity(4)
 
-    # compute T_we. Tuned once and used forever onward 
-    # Ensures pointer points at object with right arm
-    # TODO need to add increased documentation once resolved 
-    arm_start =  numpy.array([[ 0.07277, -0.71135,  0.69905,  0.58575],
-                              [ 0.03360,  0.70226,  0.71111, -0.44275],
-                              [-0.99678, -0.02825,  0.07501,  0.95686],
-                              [ 0.     ,  0.     ,  0.     ,  1.     ]])
-		
-    T_base_obj = numpy.array([[1, 0, 0, 1.0],
-                              [0, 1, 0, 0.0],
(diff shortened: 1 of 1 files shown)


src/herbpy/tsr/fuze.py
```diff
@@ -0,0 +1,55 @@
...
+        with manip.GetRobot():
+            manip.SetActive()
+            manip_idx = manip.GetRobot().GetActiveManipulatorIndex()
+		
+    # compute T_ow
+    T0_w_0 = bottle.GetTransform()
+    T0_w_1 = numpy.identity(4)
+
+    # compute T_we. Tuned once and used forever onward 
+    # Insures pointer points at object with right arm
```
mkoval: Does this actually only work for the right arm? If so, you should check which manipulator was passed as an argument. If not, please update the comment.

src/herbpy/tsr/fuze.py
```diff
@@ -0,0 +1,55 @@
...
+from prpy.tsr.tsr import *
+
+@TSRFactory('herb', 'fuze_bottle', 'point')
+def point_at_obj(robot, bottle, manip=None):
+    '''
+    @param robot The robot performing the point
+    @param bottle The bottle to point
+    @param manip The manipulator to perform the grasp. if None 
+        the active manipulator on the robot is used
+    '''
```
mkoval: Two typos:

- "The bottle to point" -> "The bottle to point at"
- "to perform the grasp" -> "to point with"

Also, I generally see docstrings with double quotes (`"""`) instead of single quotes (`'''`).

src/herbpy/tsr/fuze.py
rachelholladay: This was a tad careless of me. 

This actually only works on the right arm. Pointing from anything other then an arm doesnt make sense and this is actually tuned to right arm. What's the stylistically koval-approved way for setting it to be the right arm?

mkoval: This action is in HerbPy (not PrPy), so I think it's fine to check `manip.GetName() == 'right'`. If that check fails, raise an exception.

mkoval:
Overall, looks good. I'll merge the pull request as soon as you fix those issues. You can update the pull request by pushing more changes to the `feature/pointing_tsr` branch.

src/herbpy/tsr/fuze.py
```diff
@@ -0,0 +1,55 @@
...
+            manip.SetActive()
+            manip_idx = manip.GetRobot().GetActiveManipulatorIndex()
+		
+    # compute T_ow
+    T0_w_0 = bottle.GetTransform()
+    T0_w_1 = numpy.identity(4)
+
+    # compute T_we. Tuned once and used forever onward 
+    # Insures pointer points at object with right arm
+    arm_start =  numpy.array([[ 0.07277, -0.71135,  0.69905,  0.58575],
```
siddhss5: You need to be more specific about what this array actually is. `arm_start` has no meaning. Is it a transform? Maybe call it `Tarm_start` then. I actually really prefer a `dict` like `T["arm_start"]`.

src/herbpy/tsr/fuze.py
```diff
@@ -0,0 +1,55 @@
...
+       			      [ 0.03360,  0.70226,  0.71111, -0.44275],
+    		              [-0.99678, -0.02825,  0.07501,  0.95686],
```
siddhss5: Needs to be formatted.

src/herbpy/tsr/fuze.py
```diff
@@ -0,0 +1,55 @@
...
+ 			      [ 0.     ,  0.     ,  0.     ,  1.     ]])
+		
+    base_obj_trans = numpy.array([[1, 0, 0, 1.0],
```
siddhss5: Again, inconsistent naming. Stick to `T` or `_trans`. Don't switch back and forth.

src/herbpy/tsr/fuze.py
```diff
@@ -0,0 +1,55 @@
...
+                                  [0, 1, 0, 0.0],
+                                  [0, 0, 1, 1.0],
+                                  [0, 0, 0, 1.0]])
+
+    TW_e_0 = numpy.dot(numpy.linalg.inv(base_obj_trans), arm_start)
```
siddhss5: No idea what's going on here. Can you add comments in the document?

src/herbpy/tsr/fuze.py
siddhss5: I'm afraid this is super-brittle. I'm guessing this is the relative transform from the hand base frame to the tip of the finger that is pointing. One major issue is that this assumes that the hand is already preshaped to the desired shape, a shape that is not part of this function at all. 
My suggestion is to create an empty link in the URDF and give it a descriptive name like `pointer` and just get its relative transform. It's also quite generalizable. What do you think @mkoval?

src/herbpy/tsr/fuze.py
siddhss5: Are you really trying to `insure`? Or 'ensure`?

src/herbpy/tsr/fuze.py
```diff
@@ -0,0 +1,55 @@
+import numpy
+from prpy.tsr.tsrlibrary import TSRFactory
+from prpy.tsr.tsr import *
```
siddhss5: Stylistically, I'm not a fan of `from X import Y`. Even much less of a fan of `from X import *`. I'd rather have `import X` and write out explicitly in the code. Mostly a nitpick.

src/herbpy/tsr/fuze.py
mkoval: I would prefer to not add a bogus link to the URDF. It will just add more clutter to our already quite cluttered TF tree.

We have have the `/right/finger0_2`, `/right/jfinger1_2`, and `/right/finger2_2` links. The origin of these links is aligned with one axis along the axis of revolution of the distal joint and another axis pointed towards the fingertip. We should only have to apply a small offset along one axis to get the position of the fingertip.

I agree that this assumes that the hand is already in a sane preshape. We could encapsulate changing the hand preshape and planning to this TSR using @jeking04's `ActionLibrary`. See #15 for an example.

mkoval: I would interpret` Tarm_start` as a "transform from the `start` frame to the `arm` frame. Is that actually the meaning of this transform? What is the `arm` frame? What is the `start` frame?

@siddhss5 I haven't see the `dict` convention before; e.g. `T['arm_start']`. Why is that better than just using a `T` prefix like in `Tarm_start`? It seems to make the syntax a bit cluttered.

src/herbpy/tsr/fuze.py
mkoval: I agree entirely with `from X import *`. This is going to cause us pain down the road when we get a name conflict.

I'm generally fine with `from X import Y`, but prefer them to be in the function namespace. In this case, I don't see any harm in moving all of the imports inside the function definition.

src/herbpy/tsr/fuze.py
siddhss5: I'm cool with a prefix, as long as it's consistent. Nice thing about the dict is that you can then do ensemble things with it if needed.

src/herbpy/tsr/fuze.py
jeking04: This is from me setting a bad example. Rachel can you fix it here and I'll fix it in the other files (bowl, plate, etc)?

src/herbpy/tsr/fuze.py
rachelholladay: I'm not sure how much of the above conversation this invalidates but arm_start isn't relating to the fingers and hand. I should have been more clear in my documentation. I'll explain (the slight hack) that's going on. If you all have a better method I'll gladly use it, otherwise I'll add this documentation in:

For pointing the arm has to pointing directly at the object*. So what I did was place an object at an arbitrary location (base_obj_trans) and manually moved the arm around in till it was making a direct point. The arm's transform resulting from this tuning is 'arm_start'. Setting this to be the TW_e basically sets it such that our arm points directly at the object. The TSR and specifically Bw allow the pointer to move in a sphere around the object where that sphere is of varying radius (hence this corresponds to begin closer or further from the object and pointing at it from a different angle).

Something that makes me sad is that when this gets generalized to say, ADA, there will be this initial calibration step where you define what it means for an arm to pointing directly at an object. The only slight comfort is that you should
only have to do this once at the very beginning.

This was the method Jen and I came up with for ensuring the pointer is actually
pointing directly at the object (because that what you want to end state to be).
 If another method comes to mind, that would be totally neat.

The act of curling the fingers to make the canonical pointing preshape for the f
ingers and hand would totally be handled in the action library. Right now its be
ing treated as like a post-processing step.

*caveat: One could say that exaggeration improves legibility. However whether this is true or not is super dependent on your cost functions. For example, for the cost functions used in the ROMAN paper exaggerating slightly in the direction of the goal decreased legibility. I also personally think exaggeration makes less sense wrt to pointing, so I'm going to discount it.

siddhss5: That really makes absolutely no sense.

rachelholladay: Can you clarify what was confusing? Alternatively it'll be much easier to explain this in person.

src/herbpy/tsr/fuze.py
jeking04: Ok, I think a lot of confusion would be mitigated by just performing this multiplication:
```numpy.dot(numpy.linalg.inv(base_obj_trans), arm_start)```
once offline and then putting the result in here.  So instead of arm_start or base_obj_trans you just have:

```
Tw_e_0 = numpy.array([[ 0.07277, -0.71135,  0.69905, -0.41425],
       [ 0.0336 ,  0.70226,  0.71111, -0.44275],
       [-0.99678, -0.02825,  0.07501, -0.04314],
       [ 0.     ,  0.     ,  0.     ,  1.     ]])
```
Also, we should try this for the left arm too.  I can't quite remember how the two end-effector frames are setup relative to the hands, but there is a chance this is valid for both - if each hand is oriented the same relative to its associated end-effector transform.

File: src/herbpy/tsr/fuze.py [hunk 8]
```diff
@@ -0,0 +1,54 @@
...
+import prpy.tsr 
+
+@prpy.tsr.tsrlibrary.TSRFactory('herb', 'fuze_bottle', 'point')
+def point_at_obj(robot, bottle, manip='right'):
+    '''
+    @param robot The robot performing the point
+    @param bottle The bottle to point at
+    @param manip The manipulator to point with. This must be the right arm 
+    '''
+    if manip.GetName() == 'right':
```
mkoval: This isn't right. You probably want something like this:

```python
def point_at_obj(robot, bottle, manip=None):
    if manip is None:
        manip = robot.right_arm
    if manip.GetName() != 'right':
        raise prpy.exceptions.PrPyException('Pointing is only defined on the right arm.')
    # ...
```

PR #15, opened 2015-03-29
Feature/action library

jeking04:
This is tied with a pull request for a branch of the same name in prpy. If we decide this looks good, the prpy pull request should be merged first.

### DIFF
--- a/src/herbpy/action/grasping.py
+++ b/src/herbpy/action/grasping.py
@@ -29,7 +29,9 @@ def PushGrasp(robot, obj, distance=0.1, manip=None, preshape=[0., 0., 0., 0.], p
     """
     HerbGrasp(robot, obj,  manip=manip, preshape=preshape, distance=distance)
 
-def HerbGrasp(robot, obj, distance=None, manip=None, preshape=[0., 0., 0., 0.], push_required=False,
+def HerbGrasp(robot, obj, distance=None, manip=None, 
+              preshape=[0., 0., 0., 0.], 
+              push_required=False, 
               **kw_args):
     """
     @param robot The robot performing the push grasp
@@ -37,19 +39,16 @@ def HerbGrasp(robot, obj, distance=None, manip=None, preshape=[0., 0., 0., 0.], 
     @param distance The distance to push before grasping (if None, no pushing)
     @param manip The manipulator to perform the grasp with 
        (if None active manipulator is used)
+    @param preshape The grasp preshape for the hand
     @param push_required If true, throw exception if a plan for the pushing 
        movement cannot be found. If false, continue with grasp even if push 
        cannot be executed. (only used if distance is not None)
-    @param preshape The grasp preshape for the hand
     """
     if manip is None:
         manip = robot.GetActiveManipulator()
 
     # Move the hand to the grasp preshape
-    manip.hand.MoveHand(f1 = preshape[0],
-                        f2 = preshape[1],
-                        f3 = preshape[2],
-                        spread = preshape[3])
+    manip.hand.MoveHand(*preshape)
 
     # Get the grasp tsr
     grasp_tsr_list = robot.tsrlibrary(obj, 'grasp') 
--- a/src/herbpy/action/grasping.py
+++ b/src/herbpy/action/grasping.py
@@ -40,7 +40,7 @@ def HerbGrasp(robot, obj, distance=None, manip=None, preshape=[0., 0., 0., 0.], 
     manip.hand.MoveHand(f1 = preshape[0],
                         f2 = preshape[1],
                         f3 = preshape[2],
-                        f4 = preshape[3])
+                        spread = preshape[3])
 
     # Get the grasp tsr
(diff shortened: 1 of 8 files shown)


src/herbpy/action/grasping.py
```diff
@@ -0,0 +1,93 @@
...
+    @param preshape The grasp preshape for the hand
+    """
+    if manip is None:
+        manip = robot.GetActiveManipulator()
+
+    # Move the hand to the grasp preshape
+    manip.hand.MoveHand(f1 = preshape[0],
+                        f2 = preshape[1],
+                        f3 = preshape[2],
+                        spread = preshape[3])
```
mkoval: It would be cleaner to call `manip.hand.MoveHand(*preshape)` here.

src/herbpy/action/grasping.py
```diff
@@ -0,0 +1,93 @@
...
+        active_manip = robot.GetActiveManipulator()
+        robot.SetActiveManipulator(manip)
+        robot.Grab(obj)
+        robot.SetActiveManipulator(active_manip)
+                
+        # Now execute the straight line movement
+        ee_in_world = manip.GetEndEffectorTransform()
+        push_direction = ee_in_world[:3,2]
+        with prpy.viz.RenderVector(ee_in_world[:3,3], push_direction,
+                                   distance, robot.GetEnv()):
```
mkoval: I like the fact that we render debug information by default. It would, however, be nice to have a way to disable this rendering when we're calling these functions in the background; e.g. using Evan's planner. Maybe add a `render` argument that defaults to `True`?

src/herbpy/tsr/fuze.py
```diff
@@ -0,0 +1,33 @@
+import numpy
+from prpy.tsr.tsrlibrary import TSRFactory
+from prpy.tsr.tsr import *
+
+@TSRFactory('herb', 'fuze_bottle', 'grasp')
+def fuze_grasp(robot, fuze, manip=None):
```
mkoval: Is this independent of the rest of the pull request? When would we want to use `TSRLibrary` to perform a grasp on a Fuze bottle versus the push-grasp?

src/herbpy/action/grasping.py
jeking04: Done

src/herbpy/action/grasping.py
jeking04: I like this idea. Any tips on the cleanest way to do this? Should I do something like:
```
if render:
    with ...:
         robot.Plan...
else:
   robot.Plan...
```

Or assign a dummy function to get called by the with block when render is False:
```
if render:
   f = prpy.viz.Render....
else
  f = lambda x,y,z: pass

with f(...):
   robot.Plan...
```

Or is there a completely different option that is better?

src/herbpy/tsr/fuze.py
jeking04: The TSRFactory function just returns the list of TSRChains that can be used to successfully complete a grasp. So this function is actually used by the PushGrasp action. In PushGrasp we call ```robot.tsrlibrary(object, 'grasp')```  to get the TSR to plan to. We hadn't yet defined any TSRs for the fuze bottle.

src/herbpy/action/grasping.py
mkoval: I ran into this before and never came up with a satisfactory answer. Since this is a recurring problem, maybe we should add a `render` flag to our visualization context managers (including `RenderTSRList` and `RenderVector`). If `not render`, then the context manager is a no-op.

It is, however, a bit strange to call `RenderVector(x, render=False)`. But I don't see a better option?

src/herbpy/tsr/fuze.py
mkoval: I see, that makes sense. In that case, does the `TSRFactory` return a grasp with the end-effector directly adjacent to the bottle?

If so, it seems like `PushGrasp` should add the standoff distance to the beginning of the `TSRChain`. Otherwise, assuming a roughly symmetric distribution, there is a 50% chance that the wrist will hit the object we're trying to grasp during the `PlanToTSR` call.

mkoval: I am thinking something like this:

```python
# Get the grasp tsr
grasp_tsr_list = robot.tsrlibrary(obj, 'grasp') 

# Add the standoff distance.
Tstandoff = numpy.array(
    [ 1., 0., 0., 0. ],
    [ 0., 1., 0., 0. ],
    [ 0., 0., 1., standoff_distance ],
    [ 0., 0., 0., 1. ],
])
for tsr_chain in grasp_tsr_list:
    tsr_chain.append(TSR(T0_w=Tstandoff, Tw_e=numpy.eye(4), Bw=numpy.zeros((6, 2)), manip=manip_idx))
```
(This is untested pseudo-code, so proceed with caution).

jeking04: Okay I think we have this addressed properly now. The PushGrasp action takes a push_distance parameter and passes that onto the TSRlibrary which can use it or ignore it as appropriate.

PR #16, opened 2015-04-02
cleaning up import style

rachelholladay:
Since (as mentioned when doing fuze.py) the preferred method is to explicitly list libraries instead of using "from prpy.tsr.tsr import *" or "from prpy.tsr.tsrlibrary import TSRFactory" I made that correction across the rest of the library.


rachelholladay:
Deleted in favor of more updated pull request at #38

PR #17, opened 2015-04-02
Updated the default planning pipeline

mkoval:
This makes use of the new `MethodMask` and `FirstSupported` meta-planners in PrPy. See https://github.com/personalrobotics/prpy/pull/81 for more information.

@siddhss5 What do you think of this logic as the default?


mkoval:
Merging this so @jeking04 can test it in the demo today. It is easy to change this behavior in the future, if necessary.

PR #30, opened 2015-05-25
correction to tsr values for pointing tsr

rachelholladay:
Correction to constants of the second Bw of the pointing tsr, they were a little off and therefore caused the bug of being able to point in the exact opposite direction of an object


mkoval:
I trust @rachelholladay's judgement here. She's the pointing expert. :smile:

PR #36, opened 2015-08-12
remove rogue from branch, fix permissions

rachelholladay:
Removed feature/rogue since it's a separate branch. Fixed file permissions.


mkoval:
This is replaced by #37.

PR #38, opened 2015-08-13
Feature/tsr import style

rachelholladay:
Systematic clean up of the import style across all files in the tsr folder. Removes the need for `from __ import *` that made people cringe.


rachelholladay:
When this is merged in please also delete the branch `feature/import_corrections` and its subsequent pull request in favor of this one. I lack the permission to do so.

mkoval:
Thanks!

For future reference, we don't have a problem with `from x import y`. This can make for much less verbose code if used judiciously. We're only trying to avoid `from x import *`, which is a recipe for all sorts of unexpected name conflicts.

PR #39, opened 2015-08-13
add talker robot.Say() function

ClintLiddick:
Works on herb and in simulation


psigen:
So clean! :+1:

ClintLiddick:
@psigen @mkoval, @mharding01 reminded me of a good point. We can only launch one ROS node per process, so is this really the right way to do this? Should we be starting a single master "herb_node" that can call different services, rather than starting a "talker_client" node?

mkoval:
Yes, we can only start one node per process. In C++, e.g. [here in `or_owd_controller`](https://github.com/personalrobotics/or_owd_controller/blob/master/src/OWDControllerPlugin.cpp#L63) we handle this by starting a new node if and only if one does not already exist. We also pass the `AnynomousNode` flag to avoid name conflicts when multiple instances of HerbPy are connected to the same `roscore` instance.

I believe the C++ and Python client libraries are completely independent, so we'll actually have two ROS nodes per HerbPy process: one in C++ and one in Python. This shouldn't be an issue as long as both have unique names and/or are initialized with the `AnonymousName` flag.

**TL;DR** I think the initialization code should look something like this:

```python
if not self.talker_simulated:
    if not rospy.core.is_initialized():
        rospy.init_node('herbpy', anonymous=True)
        logger.debug('Started ROS node with name "%s".', rospy.get_name())

    # Create the actionlib client.
```

src/herbpy/herbrobot.py
```diff
@@ -229,3 +237,20 @@ def DetectObjects(self,
...
+    def Say(self, words, block=True):
+        """Speak 'words' using talker action service or espeak locally in simulation"""
+        if self.talker_simulated:
+            import subprocess
+            try:
+                proc = subprocess.Popen(['espeak', '-s', '160', '"{0}"'.format(words)])
+                if block:
+                    proc.wait()
+            except OSError:
+                logger.error('Unable to speak. Make sure "espeak" is installed locally.')
```
mkoval: You should also log the `OSError` message text.

src/herbpy/herbrobot.py
```diff
@@ -229,3 +237,20 @@ def DetectObjects(self,
         except Exception, e:
             logger.error('Detection failed update: %s' % str(e))
             raise
+
+    def Say(self, words, block=True):
+        """Speak 'words' using talker action service or espeak locally in simulation"""
+        if self.talker_simulated:
+            import subprocess
+            try:
+                proc = subprocess.Popen(['espeak', '-s', '160', '"{0}"'.format(words)])
```
mkoval: Please add `espeak` as an `<exec_depend>` in the `package.xml` file.

ClintLiddick:
@mkoval check out `herbrobot.py` for changes.

Also, I'm now getting build errors when I try to add `<exec_depend>` to package.xml because it's not format 2. I'll submit a separate pull request for converting to format 2.

src/herbpy/herbrobot.py
```diff
@@ -173,10 +173,22 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
         import herbpy.action
         import herbpy.tsr
 
+        # Initialize herbpy ROS Node
+        import rospy
+        if not rospy.core.is_initialized():
+            rospy.init_node('herbpy', anonymous=True)
+            logger.debug('Started ROS node with name "%s".', rospy.get_name())
+
```
mkoval: Please move this back into the `if not self.talker_simulated` block. Starting a ROS node is only required if you want to use the talker service, so there's no point in spinning on up by default.

mkoval:
@ClintLiddick It looks like you'll need to merge in #40 and fix a conflict in `package.xml`. However, this should make the `<exec_depend>` you added work. :)

ClintLiddick:
@mkoval got it. I also incremented the herbpy minor version, since it's a non-breaking api change. Is that how you guys do it? Let me know.

mkoval:
The version number is bumped as part of the [release process](https://wiki.personalrobotics.ri.cmu.edu/doku.php?id=documentation:creating_a_tag). You shouldn't edit it manually.

I'll merge the pull request as soon as you remove the change to the version number.

ClintLiddick:
Ahh, ok. Backed out the version change.

ClintLiddick:
Ok, should finally be ready.

jeking04:
What do you say @mkoval? can we merge this?

mkoval:
Yup! I didn't realize that @clint fixed the remaining issues.

ClintLiddick:
Wonderful! Thanks! @rachelholladay and I used it this afternoon

PR #42, opened 2015-09-16
Feature/demo rogue

rachelholladay:
Adding the gesture interface for the demo. This involves:
- some pre-computed trajectories for waving, stored in config/
- a slightly simplified version of gestures in the action lib
- the generic tsrs for gestures in tsr lib
- a script file for a text interface to control the gestures


mkoval:
@jeking04 This is a pull request into the `block_sorting` branch. Could you take a look at it and merge it if it looks good to you?

mkoval:
I think this was merged as part of #48. Please re-open this if that is not the case.

rachelholladay:
It wasn't. I supposed it was ignored. I'll PR to master.

PR #43, opened 2015-09-17
Switched to blending-only for constrained trajectories

mkoval:
This pull request disables short-cutting for constrained trajectories. I enabled blending with some heuristic parameters that seem to produce relatively smooth trajectories that follow the constraint.

This is a temporary workaround because we don't have a proper constrained smoothing.

### DIFF
--- a/src/herbpy/herbrobot.py
+++ b/src/herbpy/herbrobot.py
@@ -143,10 +143,9 @@ class HERBRobot(Robot):
         )
         
         from prpy.planning.retimer import HauserParabolicSmoother
-        self.smoother = HauserParabolicSmoother()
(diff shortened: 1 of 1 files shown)


mkoval:
@jeking04 Thoughts on this? If you don't mind, could you try running the demo with these changes and see if you notice any difference?

mkoval:
@jeking04 Could you give this a shot and see how the trajectories look?

jeking04:
@mkoval, I ran several runs of the demo yesterday on HERB with this branch and everything looked good.

mkoval:
If it's good enough for @jeking04, it's good enough for me. Merging!

PR #53, opened 2015-10-13
Bugfix/issue52

jeking04:
This adds logic to prevent herbpy from crashing when pr-ordata is not in the workspace.

### DIFF
--- a/src/herbpy/herbrobot.py
+++ b/src/herbpy/herbrobot.py
@@ -179,20 +179,27 @@ class HERBRobot(Robot):
         self.segway_sim = segway_sim
 
         # Set up perception
+        self.detector=None
         if perception_sim:
             from prpy.perception import SimulatedPerceptionModule
             self.detector = SimulatedPerceptionModule()
         else:
             from prpy.perception import ApriltagsModule
-            kinbody_path = prpy.util.FindCatkinResource('pr_ordata',
-                                                        'data/objects')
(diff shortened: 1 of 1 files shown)


src/herbpy/herbrobot.py
```diff
@@ -179,20 +179,27 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
...
+                self.detector = ApriltagsModule(marker_topic='/apriltags_kinect2/marker_array',
+                                                marker_data_path=marker_data_path,
+                                                kinbody_path=kinbody_path,
+                                                detection_frame='head/kinect2_rgb_optical_frame',
+                                                destination_frame='map')
+            except IOError as e:
+                 logger.warning('Failed to find required resource path. ' \
+                                'pr-ordata package cannot be found. ' \
+                                'Perception detector will not be loaded.')
+
```
mkoval: Is it better to load nothing or simulated perception in this case?

jeking04: I think it would be misleading to load simulated perception when we don't pass the ```perception-sim``` flag.  If perception is needed, its likely not installing pr-ordata was an oversight by the developer. Setting the detector to ```None``` will likely cause errors in calling scripts that use perception, allowing the oversight to easily be detected and fixed.

mkoval: :+1:

PR #54, opened 2015-10-15
Added a gesture interface

rachelholladay:
Adding the gesture interface for the demo. This involves:

- some pre-computed trajectories for waving, stored in config/
- a slightly simplified version of gestures in the action lib
- the generic tsrs for gestures in tsr lib
- a script file for a text interface to control the gestures


src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+    elif (type(focus) == numpy.ndarray) and (focus.ndim == 1):
+        if len(focus) != 3:
+            raise prpy.exceptions.PrPyExceptions('A point in space must \
+                    contain exactly three coordinates')
+        focus_trans = numpy.eye(4)
+        focus_trans[0:3, 3] = focus
+        goal_name = None
+    else:
+        raise prpy.exceptions.PrPyException('Focus of the point is an \
+                unknown object')
```
mkoval: This type inference is un-Pythonic and makes me nervous. For example, the "pointing at a point in space" logic will break if the user passes a `list`, `tuple`, or any other type of collection that is not an `ndarray`. This violates Python's affinity for [duck typing](https://en.wikipedia.org/wiki/Duck_typing).

I would prefer if you restructured this slightly:

1. Make a `Point` action that takes `focus_trans` and (maybe, see above) a `goal_name`.
2. Add `PointAtKinBody` and `PointAtPoint` functions for, respectively points and `KinBody`'s
3. If you think it is necessary, write a generic `PointAt` function that infers which `PointAtX` function to call.

Also, you should [always use `instanceof`, not `type` equality checks](http://legacy.python.org/dev/peps/pep-0008/#programming-recommendations) to compare types.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+                 that is being pointed at
+    @param manip The manipulator to perform the point with. 
+                 This must be the right arm
+    @param render Render tsr samples during planning
+    """
+
+    #Pointing at an object
+    if type(focus) == openravepy.openravepy_int.KinBody:
+        focus_trans = focus.GetTransform()
+        goal_name = focus.GetName()
```
mkoval: What is the purpose of `goal_name`? It does not seem to be used anywhere.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+        focus_trans[0:3, 3] = focus
+        goal_name = None
+    else:
+        raise prpy.exceptions.PrPyException('Focus of the point is an \
+                unknown object')
+
+    if manip is None:
+        print "Setting the right arm to be the active manipulator."
+        manip = robot.right_arm
+        manip.SetActive()
```
mkoval: This changes the state of the active manipulator. You should wrap this, and any code that relies on the manipulator being active, in a `RobotStateSaver` with the appropriate flags set.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+    @param focus The 3-D coordinate in space or object 
+                 that is being pointed at
+    @param manip The manipulator to perform the point with. 
+                 This must be the right arm
+    @param render Render tsr samples during planning
+    """
+
+    #Pointing at an object
+    if type(focus) == openravepy.openravepy_int.KinBody:
+        focus_trans = focus.GetTransform()
```
mkoval: You need to lock the environment around `GetTransform`.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+    #Presenting a point in space
+    elif (type(focus) == numpy.ndarray) and (focus.ndim == 1):
+        if len(focus) != 3:
+            raise prpy.exceptions.PrPyExceptions('A presentation in space \
+                    must contain exactly three coordinates')
+        focus_trans = numpy.eye(4)
+        focus_trans[0:3, 3] = focus
+    else:
+        raise prpy.exceptions.PrPyException('Focus of the presentation is an \
+                unknown object')
```
mkoval: See above. My same comments apply here.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+
+    if manip is None:
+        print "Setting the right arm to be the active manipulator."
+        manip = robot.right_arm
+        manip.SetActive()
```
mkoval: Same as above. Need a `RobotStateSaver`.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+    elif (type(end) == numpy.ndarray) and (end.ndim == 1):
+        if len(end) != 3:
+            raise prpy.exceptions.PrPyExceptions('A point in space must \
+                    contain exactly three coordinates')
+        end_coords = end
+        end_trans = numpy.eye(4)
+        end_trans[0:3, 3] = end_coords
+    else:
+        raise prpy.exceptions.PrPyException('Focus of the point is an \
+                unknown object')
```
mkoval: Same as above. In fact, there is a lot of code duplication here. I wonder if you could factor out this logic into a helper function that contains the switch-case statement for input types?

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+    @param distance The distance the object will be lifted up
+    @param wait The amount of time the object will be held up in seconds
+    @param render Render tsr samples during planning
+    """
+
+    if manip is None:
+        manip = robot.GetActiveManipulator()
+
+    preconfig = manip.GetDOFValues()
+    robot.Grasp(obj)
```
mkoval: Need to lock around `GetActiveManipulator`, `GetDOFValues`, and `Grasp`.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+    time.sleep(wait)
+
+    #'Unlift' the object, so place it back down
+    unlift_tsr = robot.tsrlibrary(obj, 'lift', manip, distance=-distance)
+    
+    with prpy.viz.RenderTSRList(unlift_tsr, robot.GetEnv(), render=render):
+        robot.PlanToTSR(unlift_tsr, execute=True)
+
+    if release:
+        robot.Release(obj)
```
mkoval: Need to lock around `Release`.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+    elif word == 'yes':
+        for i in xrange(inc):
+            robot.head.Servo([ 0,  1])
+            time.sleep(pause)
+        for i in xrange((inc*3)):
+            robot.head.Servo([ 0, -1])
+            time.sleep(pause)
+        for i in xrange((inc*2)):
+            robot.head.Servo([ 0,  1])
+            time.sleep(pause)
```
mkoval: Why is this implemented as a sequence of `Servo` commands, instead of a trajectory?

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+    with prpy.viz.RenderTSRList(unlift_tsr, robot.GetEnv(), render=render):
+        robot.PlanToTSR(unlift_tsr, execute=True)
+
+    if release:
+        robot.Release(obj)
+        manip.hand.OpenHand()
+        manip.PlanToConfiguration(preconfig)
+
+@ActionMethod
+def Nod(robot, word='yes', inc=4):
```
mkoval: The `yes` and `no` cases have little in common. Can you split this into two functions? Perhaps `NodHead` and `ShakeHead`? Or maybe `NodYes` and `NodNo`?

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+                             1.46215605, -0.12918424])
+
+        manip.PlanToConfiguration(pose)
+        robot.left_hand.MoveHand(f1=0, f2=0, f3=0, spread=3.14)
+    else: 
+        raise prpy.exceptions.PrPyException('Stop is only defined \
+                                                for the left and right arm.')
+
+@ActionMethod
+def MiddleFinger(robot, manip=None):
```
mkoval: lol. :laughing:

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+    manip = robot.right_arm
+    manip.SetActive()
+    print "Setting right arm as the active manipulator"
+
+    wave_path = FindCatkinResource('herbpy', 'config/waveTrajs/')
+    try:
+        traj0 = load_trajectory(env, wave_path+'wave0.xml')
+        traj1 = load_trajectory(env, wave_path+'wave1.xml')
+        traj2 = load_trajectory(env, wave_path+'wave2.xml')
+        traj3 = load_trajectory(env, wave_path+'wave3.xml')
```
mkoval: You should use `os.path.join` here, not string concatenation.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,321 @@
...
+    except IOError as e:
+        raise ValueError('Failed loading wave trajectory from "{:s}".'.format(
+            wave_path))
```
mkoval: Raise an `IOError` here, not a ValueError`. Also, please include the error message `str(e)` of the original exception in the one you create.

src/herbpy/tsr/generic.py
```diff
@@ -0,0 +1,156 @@
...
+def point_obj(robot, transform, manip=None):
+    """
+    @param robot The robot performing the point
+    @param transform The location of where the robot is pointing to
+    @param manip The manipulator to point with. This must be the right arm. 
+    """
+    if manip is None:
+        manip = robot.right_arm
+
+    with manip.GetRobot():
```
mkoval: I would prefer if this created a `RobotStateSaver` that only preserved the active DOFs and active manipulator, instead of using the default options.

mkoval: This applies to the other `TSRFactory`s in this file, too.

mkoval:
Overall, this looks pretty good. I left a number of relatively minor comments, mostly about missing locks and `RobotStateSaver`s.

src/herbpy/action/rogue.py
rachelholladay: Suggestion from @mklingen. It also works better in practice (having tested both).

src/herbpy/action/rogue.py
rachelholladay: Can you clarify what you mean by that or point me to an example where that's done? I don't know how to do this..

mklingen:
````
with env:
  #do stuff
```

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,354 @@
...
+    @param robot The robot performing the point
+    @param focus The 3-D coordinate in space or object 
+                 that is being pointed at
+    @param manip The manipulator to perform the point with. 
+                 This must be the right arm
+    @param render Render tsr samples during planning
+    """
+    env = robot.GetEnv()
+    pointing_coord = GetPointFrom(env, focus)
+    Point(robot, pointing_coord, manip, render)
```
mkoval: You should propagate the `return` value; i.e.
```python
return Point(robot, pointing_coord, manip, render)
```

mkoval: This applies to all of the wrapper functions.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,354 @@
...
+def GetPointFrom(env, focus):
+    """
+    @param env The environment where the item exists
+    @param focus The area to be referred to
+    """
+    #Pointing at an object
+    if isinstance(focus, openravepy.KinBody):
+        with env: 
+            focus_trans = focus.GetTransform()
+        coord = list(focus_trans[0:3, 3])
```
mkoval: You should also support `openravepy.KinBody.Link`. I think the same code will work, so you can simply change the condition to:

```yaml
if isinstance(focus, (openravepy.KinBody, openravepy.KinBody.Link)):
```

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,354 @@
...
+
+    #Pointing at a point in space as numpy array
+    elif (isinstance(focus, numpy.ndarray) and (focus.ndim == 1) 
+           and (len(focus) == 3)):
+        coord = list(focus)
```
mkoval: You should also add support for a 4 x 4 transformation matrix.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,354 @@
...
+
+    #Pointing at a point in space as list
+    elif isinstance(focus, list):
+        coord = focus
+
+    #Pointing at a point in space as tuple
+    elif isinstance(focus, tuple):
+        coord = list(focus)
```
mkoval: You should be able to combine the `list` and `tuple` cases:
```python
elif isinstance(focus, (list, tuple)):
    coord = list(focus)
````
Even better, just check if `len(x) == 3`.

File: src/herbpy/action/rogue.py [hunk 20]
```diff
@@ -0,0 +1,354 @@
...
+
+    else:
+        raise prpy.exceptions.PrPyException('Focus of the point is an \
+                unknown object')
+
+    return coord
+
+def Point(robot, coord, manip=None, render=False):
```
mkoval: Shouldn't this be an `@ActionMethod`?

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,354 @@
...
+
+    #ee_offset : such that the hand, not wrist, is above the object
+    #hand_pose : places the hand above the start location
+    if manip.GetName() == 'right':
+        hand = robot.right_hand
+        ee_offset = -0.2
+        hand_pose = numpy.array([[ 0, -1, 0, start_coords[0]],
+                                 [ 0,  0, 1, (start_coords[1]+ee_offset)],
+                                 [-1,  0, 0, (start_coords[2]+margin)],
+                                 [ 0,  0, 0, 1]])
```
mkoval: You should either explicitly pass `dtype='float'` or insure that at least one literal value in the array is a floating point number (e.g. replace `0` with `0.`). Otherwise, this could create an `int` array, which can lead to all sorts of unpleasant surprises.

This applies to all of the hard-coded transformations.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,354 @@
...
+
+    if manip.GetName() != 'right':
+        raise prpy.exceptions.PrPyException('Presenting is only defined \
+                on the right arm.')
+    
+    p = openravepy.KinBody.SaveParameters
+    with robot.CreateRobotStateSaver(p.ActiveManipulator):
+        robot.SetActiveManipulator(manip)
+        with prpy.viz.RenderTSRList(present_tsr, robot.GetEnv(), render=render):
+            robot.PlanToTSR(present_tsr, execute=True)
```
mkoval: You need to `SetActiveDOFs` to `manip.GetArmIndices()` before calling `PlanToTSR`. This means that you also need to update the `RobotStateSaver` accordingly.

This applies to all of the `PlanToTSR` calls in this file

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,354 @@
...
+  
+    else:
+        raise prpy.exceptions.PrPyException('Manipulator does not have an \
+                 associated hand')
+
+    end_trans = numpy.eye(4)
+    end_trans[0:3, 3] = end_coords
+
+    hand.MoveHand(f1=1, f2=1, f3=1, spread=3.14)
+    manip.PlanToEndEffectorPose(hand_pose)
```
mkoval: You need to set the active manipulator and active DOFs here (and add a `RobotStateSaver`).

mkoval: Also, this does nothing unless you pass `execute=True`.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,354 @@
...
+        if manip is None:
+            manip = robot.GetActiveManipulator()
+        preconfig = manip.GetDOFValues()
+        robot.Grasp(obj)
+
+    #Lift the object - write more tsrs
+    lift_tsr = robot.tsrlibrary(obj, 'lift', manip, distance=distance)
+    
+    with prpy.viz.RenderTSRList(lift_tsr, robot.GetEnv(), render=render):
+        robot.PlanToTSR(lift_tsr, execute=True)
```
mkoval: Missing active DOFs, active manipulator, and `RobotStateSaver`.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,354 @@
...
+
+    #Wait for 'time'
+    time.sleep(wait)
+
+    #'Unlift' the object, so place it back down
+    unlift_tsr = robot.tsrlibrary(obj, 'lift', manip, distance=-distance)
```
mkoval: I don't know if negative `distance` values are supported here.

@jeking04 Does this make sense?

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,354 @@
...
+    elif manip.GetName() == 'left':
+        pose = numpy.array([ 1.30614268, -1.76      , -1.57063853,
+                             2.07228362,  1.23918377,
+                             1.46215605, -0.12918424])
+
+        manip.PlanToConfiguration(pose, execute=True)
+        robot.left_hand.MoveHand(f1=0, f2=0, f3=0, spread=3.14)
+    else: 
+        raise prpy.exceptions.PrPyException('Stop is only defined \
+                                                for the left and right arm.')
```
mkoval: This error message is going to contain a bunch of extra spaces. Please wrap it using string concatenation:

```python
raise prpy.exceptions.PrPyException('Stop is only defined'
    ' for the left and right arm.')
```
Or, even better, just wrap outside the string:
```python
raise prpy.exceptions.PrPyException(
    'Stop is only defined for the left and right arm.')
```

mkoval: This error message is going to contain a bunch of extra spaces. Please wrap it using string concatenation:

```python
raise prpy.exceptions.PrPyException('Stop is only defined'
    ' for the left and right arm.')
```
Or, even better, just wrap outside the string:
```python
raise prpy.exceptions.PrPyException(
    'Stop is only defined for the left and right arm.')
```

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,354 @@
...
+        traj1 = load_trajectory(env, join(wave_path, 'wave1.xml'))
+        traj2 = load_trajectory(env, join(wave_path, 'wave2.xml'))
+        traj3 = load_trajectory(env, join(wave_path, 'wave3.xml'))
+    except IOError as e:
+        raise IOError(str(e))
+
+    p = openravepy.KinBody.SaveParameters
+    with robot.CreateRobotStateSaver(p.ActiveManipulator):
+        manip = robot.right_arm
+        robot.SetActiveManipulator(manip)
```
mkoval: I don't think you need to set the manipulator as active here.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,354 @@
...
+    env = robot.GetEnv()
+
+    wave_path = FindCatkinResource('herbpy', 'config/waveTrajs/')
+    try:
+        traj0 = load_trajectory(env, join(wave_path, 'wave0.xml'))
+        traj1 = load_trajectory(env, join(wave_path, 'wave1.xml'))
+        traj2 = load_trajectory(env, join(wave_path, 'wave2.xml'))
+        traj3 = load_trajectory(env, join(wave_path, 'wave3.xml'))
+    except IOError as e:
+        raise IOError(str(e))
```
mkoval: This does nothing. If you don't want to change the error message, then just remove the `try`-`except` block entirely.

src/herbpy/action/rogue.py
rachelholladay: NO. :)

src/herbpy/action/rogue.py
rachelholladay: Well I've tested and it works so its supported on some level. I like i talked to @jeking04 about this at some point.

src/herbpy/action/__init__.py
```diff
@@ -1,2 +1,3 @@
 from grasping import PushGrasp, Grasp
+from rogue import *
```
mkoval: `import *` is evil. Could you explicitly list the functions you wish to import here?

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,359 @@
...
+    @param manip The manipulator to perform the point with. 
+                 This must be the right arm
+    @param render Render tsr samples during planning
+    """
+    if manip is None:
+        manip = robot.right_arm
+
+    if manip.GetName() != 'right':
+        raise prpy.exceptions.PrPyException('Pointing is only defined \
+                on the right arm.')
```
mkoval: This will insert a bunch of spaces into the comment. You should continue the line like this:
```python
raise prpy.exceptions.PrPyException('Pointing is only defined'
                                    'on the right arm.')
```

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,359 @@
...
+    """
+    @param robot The robot performing the presentation
+    @param focus The 3-D coordinate in space or object that 
+                 is being presented
+    @param manip The manipulator to perform the presentation with. 
+                 This must be the right arm.
+    @param render Render tsr samples during planning
+    """
+    if manip is None:
+        manip = robot.right_arm
```
mkoval: We typically default to `GetActiveManipulator` if a manipulator is not specified. I would prefer if you did the same here (and in the other functions in this file).

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,359 @@
...
+    @param manip The maniplator to perform the exhibit
+    @param distance The distance the object will be lifted up
+    @param wait The amount of time the object will be held up in seconds
+    @param render Render tsr samples during planning
+    """
+    with robot.GetEnv():
+        if manip is None:
+            manip = robot.GetActiveManipulator()
+        preconfig = manip.GetDOFValues()
+        robot.Grasp(obj)
```
mkoval: I'm not sure it makes sense to `Grasp` the object in here. I would expect that the object would already be `Grasp`ed before calling this function.

@jeking04 Thoughts?

src/herbpy/action/rogue.py
mkoval: @rachelholladay Still missing these here!

src/herbpy/action/rogue.py
mkoval: If @jeking04 is fine with it, so am I!

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,359 @@
...
+        robot.SetActiveManipulator(manip)
+        robot.SetActiveDOF(manip)
+        with prpy.viz.RenderTSRList(unlift_tsr, robot.GetEnv(), render=render):
+            robot.PlanToTSR(unlift_tsr, execute=True)
+
+    if release:
+        with robot.GetEnv():
+            robot.Release(obj)
+        manip.hand.OpenHand()
+        manip.PlanToConfiguration(preconfig)
```
mkoval: Missing active DOFs. I feel like you can just move the above `RobotStateSaver` and `SetActiveDOF`/`SetActiveManipulator` block up further and wrap all of this with it.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,359 @@
...
+    elif manip.GetName() == 'left':
+        pose = numpy.array([ 1.30614268, -1.76      , -1.57063853,
+                             2.07228362,  1.23918377,
+                             1.46215605, -0.12918424], dtype='float')
+
+        manip.PlanToConfiguration(pose, execute=True)
+        robot.left_hand.MoveHand(f1=0, f2=0, f3=0, spread=3.14)
+    else: 
+        raise prpy.exceptions.PrPyException(
+            'Stop is only defined for the left and right arm.')
```
mkoval: It should say `Halt`, not `Stop`, here. :smile:

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,359 @@
...
+
+@ActionMethod
+def MiddleFinger(robot, manip=None):
+    """
+    @param robot The robot being used to give the middle finger
+    @param manip The manipulator being used to give the middle finger.
+                 Must be either the right or left arm.
+    """
+    if manip == None:
```
mkoval: `is None`, not `== None`.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,359 @@
...
+    elif manip.GetName() == 'left':
+        left_dof = numpy.array([ 1.30614268, -1.76      , -1.57063853,  
+                                 2.07228362,  1.23918377,
+                                 1.46215605, -0.12918424], dtype='float')
+
+        manip.PlanToConfiguration(left_dof, execute=True)
+        robot.left_hand.MoveHand(f1=2, f2=2, f3=0, spread=3.14)
+    else: 
+        raise prpy.exceptions.PrPyException('The middle finger is only defined \
+                                for the left and right arm.')
```
mkoval: Incorrect string wrapping. See above.

src/herbpy/action/rogue.py
```diff
@@ -0,0 +1,359 @@
...
+    env = robot.GetEnv()
+
+    wave_path = FindCatkinResource('herbpy', 'config/waveTrajs/')
+    traj0 = load_trajectory(env, join(wave_path, 'wave0.xml'))
+    traj1 = load_trajectory(env, join(wave_path, 'wave1.xml'))
+    traj2 = load_trajectory(env, join(wave_path, 'wave2.xml'))
+    traj3 = load_trajectory(env, join(wave_path, 'wave3.xml'))
+
+    p = openravepy.KinBody.SaveParameters
+    with robot.CreateRobotStateSaver(p.ActiveManipulator):
```
mkoval: Why is this block necessary? You never change the active manipulator inside of it.

src/herbpy/tsr/generic.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+
+    with robot.GetEnv():
+        if manip is None:
+            manip = robot.right_arm
+
+        with robot.CreateRobotStateSaver(
+                robot.SaveParameters.ActiveManipulator):
+            robot.SetActiveManipulator(manip)
+            manip_idx = manip.GetRobot().GetActiveManipulatorIndex()
```
mkoval: This code is duplicated several times. Can you add a `GetManipulatorIndex` function to `util` that does this?

mkoval:
All of the major issues seem to be resolved. I just left a few comments on code cleanup.

@rachelholladay Could you also add (at least) one test to each of these functions that calls it on a very simple environment? It's fine if it's something as simple as just running each function on a Fuze bottle. We have [a tutorial for doing this on the website](https://www.personalrobotics.ri.cmu.edu/software/unit-testing) and there are some [simple examples in `or_urdf`](https://github.com/personalrobotics/or_urdf/blob/master/tests/test_urdf_loader.py).

@jeking04, @Shushman, or myself can help if you run into trouble.

src/herbpy/action/rogue.py
rachelholladay: By the definition, the exhibit gesture grabs an object, lifts it up, places it back down. So I would not expect it to be grasped.

src/herbpy/action/rogue.py
rachelholladay: I discussed this with @jeking04 at some point iirc. Considering that point and present are only defined for the right arm this seemed silly. 
If we no longer agree with that then I can say
```
if manip is None:
    manip = robot.GetActiveManipulator()

check if right arm
```

src/herbpy/tsr/generic.py
rachelholladay: Something in prpy's util that takes a robot and returns the manip_index, ya?

rachelholladay:
@mkoval I made a PR on prpy for the util function, so I think that should be resolved first. I've also made most of the changes you asked for on this.

mkoval:
Thanks. The only comment I have left is that the default should *always* be `GetActiveManipulator()`, not `right_arm`. I'd much rather keep the `Action`s consistent than save a few characters of typing. Also, the current behavior will break if/when these functions support the left arm.

rachelholladay:
Alright @mkoval. Last comment addressed.

tests/rogue_tests.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python
+PKG = 'herbpy'
+import roslib; roslib.load_manifest(PKG)
+import numpy, unittest
+import herbpy, prpy
+
+env, robot = herbpy.initialize(sim=True)
+robot.right_arm.SetActive()
+fuze = prpy.rave.add_object(env, 'fuze_bottle', 'objects/fuze_bottle.kinbody.xml')
```
mkoval: This should occur in `setUp`, not in the file-level scope.

tests/rogue_tests.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python
+PKG = 'herbpy'
+import roslib; roslib.load_manifest(PKG)
```
mkoval: This is a hold out from `rosbuild` days and is not necessary with Catkin. Please remove it.

tests/rogue_tests.py
```diff
@@ -0,0 +1,69 @@
...
+robot.right_arm.SetActive()
+fuze = prpy.rave.add_object(env, 'fuze_bottle', 'objects/fuze_bottle.kinbody.xml')
+
+class RogueTest(unittest.TestCase):
+    def setUp(self):
+        self.env, self.robot, self.fuze = env, robot, fuze
+        self.fuze_pose = numpy.eye(4)
+        self.fuze_pose[0:3, 3] = [0.8, -0.3, 0.4]
+        self.fuze.SetTransform(self.fuze_pose)
+        self.fuze.Enable(True)
```
mkoval: Once you move the initialization here, you'll need to call `env.Destroy()` in `tearDown`.

rachelholladay:
Testing addressed. Can we get it merged before class? :)

PR #57, opened 2015-10-26
Added an action for invoking the push planner

jeking04:
This adds a funciton to the action library that interfaces with the push planner for pushing objects on table.


src/herbpy/action/pushing.py
```diff
@@ -0,0 +1,83 @@
...
+    
+    # Get the current pose of the manipulator and assume we want to keep that pose throughout
+    #  the push
+    with robot.GetEnv():
+        ee_pushing_transform = manip.GetEndEffectorTransform()
+    ee_pushing_transform[:2,3] = [0., 0.] #ignore x,y pose
+
+    # Compute the goal pose
+    with robot.GetEnv():
+        goal_pose = obj.GetTransform()
```
mkoval: It would be nice to combine all of these `with env` blocks. This is just a nit, so feel free to ignore. :smile:

src/herbpy/action/pushing.py
```diff
@@ -0,0 +1,83 @@
...
+    table_height = table_pos[2] + table_extents[2] + 0.01
```
mkoval: What's the purpose of this 1 cm offset? Can you make this a parameter with a default value?

src/herbpy/action/pushing.py
```diff
@@ -0,0 +1,83 @@
...
+                                  **kw_args)
+    if traj is None:
+        raise PlanningError('Failed to find pushing plan')
+
+    # Execute
+    from prpy.viz import RenderTrajectory
+    with RenderTrajectory(robot, traj, color=[1, 0, 0, 1]):
+        if shortcut_time > 0:
+            traj = planner.ShortcutPath(timelimit=shortcut_time)
+        with RenderTrajectory(robot, traj, color=[0, 0, 1, 1]):
```
mkoval: Please add a `render` argument to this function (with default `True`, if you want), so it's possible to disable thsi visualization.

src/herbpy/action/pushing.py
```diff
@@ -0,0 +1,83 @@
...
+          if manip.simulated:
+              planner.ExecutePlannedPath()
+          else:
+              robot.ExecuteTrajectory(traj)
+              planner.SetFinalObjectPoses()
```
mkoval: What's the difference between these two cases? Why can't we use `ExecuteTrajectory` in simulation, too?

src/herbpy/action/pushing.py
```diff
@@ -0,0 +1,83 @@
...
+    @param table The table kinbody the object is resting on
+    @param goal_position The desired (x,y) position of the object in world coordinates
+    @param goal_radius The max distance from goal_position for the object to be to 
+      still consider the goal achieved
+    @param manip The manipulator to use for the push - if None the active manipulator is used
+    @param max_plan_duration The max time to run the planner
+    @param shortcut_time The amount of time to spend shortcutting, if 0. no shortcutting is performed
+    """
+    # Get a push planner
+    from or_pushing.push_planner import PushPlanner
```
mkoval: Can you catch the `ImportError` here and throw an exception with a better error message? For example, "Unable to create PushPlanner. Is the `randomized_rearrangement_planning` repository checked out in your workspace?".

mkoval:
Overall, this looks great! I'm really impressed how simple you managed to make the API given how much complexity is under the hood. I just added a few very small comments above.

@jeking04 It would also be great if you could add a simple test that calls this action. This would be useful both as a test and a usage example

src/herbpy/action/pushing.py
jeking04: We could. ```ExecutePlannedPath``` on the planner renders the object movement as well as the arm movement, basically by subsampling states (containing robot and object pose) on the original OMPL trajectory and stepping the environment through these states.  ```ExecuteTrajectory``` will just execute an arm trajectory as normal, and move the arm through all the objects. Then ```SetFinalObjectPose``` teleports the objects to their final planned locations at the end of the trajectory. I think this is our only option on the real robot, but the ```ExecutePlannedPath``` option looks a little nicer in simulation.

src/herbpy/action/pushing.py
jeking04: Done :+1:

src/herbpy/action/pushing.py
jeking04: Its not necessary. Removed.

src/herbpy/action/pushing.py
jeking04: Done. But in doing so I created a dependency on a prpy pull request getting merged first.

src/herbpy/action/pushing.py
jeking04: Done. But I created a dependency on a prpy pull request that adds the render flag to RenderTrajectory.

src/herbpy/action/pushing.py
mkoval: That makes sense. Can you add a comment that explains this to the code?

mkoval:
This looks good. Thanks for adding the example.

Could you add the example (possibly even the same example) as a test? This will let us automatically check for breakage in pr-cleanroom.

src/herbpy/action/pushing.py
jeking04: Done.

jeking04:
Okay, I took the example and made it into a test. But now I am not sure if I need to add my package (randomized_rearrangement_planning) as a dependency somewhere in the package.xml. I am guessing it should be a test_depend. Is that correct?

mkoval:
Thanks for adding those tests. The dependency you added to the `package.xml` file is correct.

I haven't forgotten about this - I'm trying to get it to build in `pr-cleanroom` before I merge the pull request.

PR #58, opened 2015-11-02
Adding Testing Call

rachelholladay:
In adding unit tests to herbpy I hadn't added the call to catkin.cmake. Now its been added (as per the instructions on the website) such that all tests are called.


mkoval:
Finally got the tests to pass. :sweat:

PR #63, opened 2015-11-24
HERB ros_control migration

ClintLiddick:
@mkoval minimal changes to match the update to prpy


ClintLiddick:
Ready for review. Need to decide if there's anything we want to merge first before tagging the final OWD version.

src/herbpy/herbpantilt.py
```diff
@@ -121,3 +122,6 @@ def FindIK(self, target):
         ik_params = openravepy.IkParameterization(target, openravepy.IkParameterization.Type.Lookat3D)
         return self.ikmodel.manip.FindIKSolution(ik_params, 0)
 
+    def GetDofValues(self):
+        """Temp override to return static position while head is immobilized"""
+        return [0, -0.3]
```
mkoval: Print a warning in here or mark this function as `@deprecated`.

src/herbpy/herbrobot.py
```diff
@@ -30,6 +34,32 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
 
         Robot.__init__(self, robot_name='herb')
 
+
+        # Controller setup
+        # TODO load controller names from same config as ros_control?
```
mkoval: I am skeptical that this is possible - we have a pretty big set of arbitrary rules about when controllers need to be loaded. If you think this is possible, please file it as an issue so we can track it.

src/herbpy/herbrobot.py
```diff
@@ -30,6 +34,32 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
...
+        self.controller_manager = None  # TODO verify never used in sim
```
mkoval: Verify that this is true and delete the `TODO` message.

src/herbpy/herbrobot.py
```diff
@@ -30,6 +34,32 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
...
+        self.controllers_always_on = []
+        controllers_manip = []
+
+        # TODO how to deal with sim on controller manager?
```
mkoval: The controller manager does not exist in simulation, so there is (hopefully) nothing to deal with. If that is the case, delete this `TODO` message.

src/herbpy/herbrobot.py
```diff
@@ -30,6 +34,32 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
...
+        self.full_controller_sim = (left_arm_sim and right_arm_sim and
+                                    left_ft_sim and right_ft_sim and
+                                    left_hand_sim and right_hand_sim and
+                                    head_sim)
+        if not self.full_controller_sim:
+            # any non-simulation requires ros and the ros_control stack
+            import rospy
+            if not rospy.core.is_initialized():
+                rospy.init_node('herbpy')
```
mkoval: I prefer to only call `init_node` in the top-level script that serves as the entry point for the application. This function may only be called once per process and takes several options that we cannot pass here.

src/herbpy/herbrobot.py
```diff
@@ -56,7 +86,54 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
...
+        # hand controllers
+        if not left_hand_sim:
+            self.controllers_always_on.append('left_hand_controller')
+
+        if not right_hand_sim:
+            self.controllers_always_on.append('right_hand_controller')
+
+        # force/torque controllers
+        if not left_ft_sim or not right_ft_sim:
+            self.controllers_always_on.append('force_torque_sensor_controller')
```
mkoval: Why don't we need to differentiate between the left and right F/T sensors here?

src/herbpy/herbrobot.py
```diff
@@ -56,7 +86,54 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
...
+        # NOTE: head is ignored until TODO new Schunk head integrated
+        if not left_arm_sim:
+            controllers_manip.append(
+                'left_gravity_compensation_controller')
+        else:
+            self.left_arm.sim_controller = self.AttachController(name=self.left_arm.GetName(),
+                                                                 args='IdealController',
+                                                                 dof_indices=self.left_arm.GetArmIndices(),
+                                                                 affine_dofs=0,
+                                                                 simulated=True)
```
mkoval: `self.AttachController` is a helper function that ignores the `args` parameter and creates an `IdealController` when `simulated` is `True`. That logic does not make sense here, so I suggest bypassing this function and creating/attaching the `IdealController` manually.

src/herbpy/herbrobot.py
```diff
@@ -56,7 +86,54 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
...
+            self.right_arm.sim_controller = self.AttachController(name=self.right_arm.GetName(),
+                                                                  args='IdealController',
+                                                                  dof_indices=self.right_arm.GetArmIndices(),
+                                                                  affine_dofs=0,
+                                                                  simulated=True)
+
+        # load and activate initial controllers
+        if self.controller_manager is not None:
+            self.controller_manager.request(self.controllers_always_on +
+                                            controllers_manip).switch()
```
mkoval: Why are `controllers_manip` separate from `self.controlers_always_on`?

src/herbpy/herbrobot.py
```diff
@@ -202,7 +279,7 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
             from prpy.perception import ApriltagsModule
             try:
                 kinbody_path = prpy.util.FindCatkinResource('pr_ordata',
-                                                            'data/objects')
+                                                            'objects')
```
mkoval: Please verify that this change works if `pr_ordata` is in the `devel`-space.

src/herbpy/herbrobot.py
```diff
@@ -212,22 +289,22 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
...
+                               'pr_ordata package cannot be found. ' \
+                               'Perception detector will not be loaded.' \
+                               '\n{}'.format(e))
 
         if not self.talker_simulated:
             # Initialize herbpy ROS Node
             import rospy
             if not rospy.core.is_initialized():
-                rospy.init_node('herbpy', anonymous=True)
+                rospy.init_node('herbpy')
```
mkoval: Same comment about `init_node` as before.

src/herbpy/herbrobot.py
```diff
@@ -242,35 +319,187 @@ def CloneBindings(self, parent):
...
+                active_controllers.append(self.base.controller)
+            else:
+                logger.warning(
+                    'Trajectory includes the base, but no base controller is'
+                    ' available. Is self.base.controller set?')
+
+        for controller in active_controllers:
+            controller.SetPath(traj)
+
+        if defer is True:
```
mkoval: We removed Trollius support from PrPy in https://github.com/personalrobotics/prpy/pull/278. It is no longer necessary to handle the `defer` argument.

src/herbpy/herbrobot.py
```diff
@@ -242,35 +319,187 @@ def CloneBindings(self, parent):
...
-                raise TrajectoryAborted()
+        value = self._ExecuteTrajectory(traj, *args, **kwargs)
+
+        # TODO meaningful to do this check here?
+        # if so can be done on value future
+        #
+        # for manipulator in active_manipulators:
+        #     status = manipulator.GetTrajectoryStatus()
+        #     if status == 'aborted':
+        #         raise TrajectoryAborted()
```
mkoval: We should replace this check with code that tests whether any of the `FollowJointTrajectory` actions failed, i..e. by inspecting the `Future` returned by the controller client.

src/herbpy/herbrobot.py
```diff
@@ -242,35 +319,187 @@ def CloneBindings(self, parent):
...
+        Stiffness False/0 is gravity compensation and stiffness True/(>0) is position
+        control.
+        @param stiffness boolean or decimal value 0.0 to 1.0
         """
-        self.head.SetStiffness(stiffness)
-        self.left_arm.SetStiffness(stiffness)
-        self.right_arm.SetStiffness(stiffness)
+        if isinstance(stiffness, numbers.Number) and not (0 <= stiffness <= 1):
+            raise Exception('Stiffness must be boolean or decimal in the range [0, 1];'
+                            'got {}.'.format(stiffness))
```
mkoval: "Decimal" is confusing here - floating point numbers are not technically decimals. :smile:

src/herbpy/herbrobot.py
```diff
@@ -242,35 +319,187 @@ def CloneBindings(self, parent):
...
+                new_manip_controllers.append('left_joint_group_position_controller')
+            if not self.right_arm.IsSimulated() and (manip is None or manip is self.right_arm):
+                new_manip_controllers.append('right_joint_group_position_controller')
+        else:
+            if not self.left_arm.IsSimulated() and (manip is None or manip is self.left_arm):
+                new_manip_controllers.append(
+                    'left_gravity_compensation_controller')
+            if not self.right_arm.IsSimulated() and (manip is None or manip is self.right_arm):
+                new_manip_controllers.append(
+                    'right_gravity_compensation_controller')
```
mkoval: How could `manip` be `None` here? It's the `self` object!

mkoval: Oh, I see. This is now a function `Robot`.

mkoval:
This is a duplicate of #24 after reverting merging that pull request into `master`. We need to merge this pull request simultaneously with https://github.com/personalrobotics/prpy/pull/233 and https://github.com/personalrobotics/herb_description/pull/25.

src/herbpy/herbrobot.py
ClintLiddick: `force_torque_sensor_controller` loads a single controller that publishes all available F/T resources. So we get the appropriate message in each individual hand's call.

src/herbpy/herbrobot.py
ClintLiddick: Same as other place, I'll leave this as-is since I would just duplicated the exact code here.

src/herbpy/herbrobot.py
ClintLiddick: Just to make it clearer what the logic blocks are dealing with

src/herbpy/herbrobot.py
ClintLiddick: #84

src/herbpy/barretthand.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def GetStrain(hand):
+        """ Gets the most recent strain sensor readings.
+        @return list of strain gauge values for each finger
+        """
+        if not hand.simulated:
+            # This is because we are overriding the force/torque sensor datatype
+            # sensor_data = hand.handstate_sensor.GetSensorData()
+            # return sensor_data.force.copy()
+            raise NotImplementedError("Strain guage not yet implemented in Python under ros_control")
```
mkoval: Typo: `strain guage`

mkoval: Also, delete dead code.

src/herbpy/wam.py
```diff
@@ -0,0 +1,359 @@
...
+        @param max_force maximum force in Newtons
+        @param max_torque maximum torque in Newton-Meters
+        @param ignore_collisions collisions with these objects are ignored when 
+        planning the path, e.g. the object you think you will touch
+        @param velocity_limit_scale A multiplier to use to scale velocity limits 
+        when executing MoveUntilTouch ( < 1 in most cases).           
+        @param **kw_args planner parameters
+        @return felt_force flag indicating whether we felt a force.
+        """
+        raise NotImplementedError('MoveUntilTouch not yet implemented under ros_control.')
```
mkoval: Only disable when not simulated.

mkoval:
Travis seems to be caching an old version of `repositories.yml` and, thus, failing to install all of HerbPy's dependencies.

I ran the unit tests locally and found that **all of the unit tests fail**. There appear to be some pretty major API breaks on the `BarrettHand` class, amongst some other issues. They also print this error message and hang indefinitely upon exit:
```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::lock_error> >'
  what():  boost: mutex lock failed in pthread_mutex_lock: Invalid argument
```

@ClintLiddick Could you investigate?

ClintLiddick:
Oh fun. I'll check it out in the morning. Most of the methods now just
throw so I guess those tests just need to be disabled for now.

mkoval:
There appears to be a transient Apt problem on Travis. I ran the tests locally and they passed.

PR #75, opened 2016-01-22
Changed DetectObjects() default params

Shushman:
Default usage of `DetectObjects()` is to call AprilTags with reference link as `herb_base`


mkoval:
This requires https://github.com/personalrobotics/prpy/pull/261.

PR #78, opened 2016-02-05
Added actions for pitcher pouring YCB task

vinitha910:
This branch contains pouring.py and actions.py. pouring.py contains the YCB pour water action and actions.py contains general actions: moving objects, push grasping a cup and grasping the pitcher.


mkoval:
It looks like Travis is unhappy because you have an old version of the `package.xml` file in your branch. 

@vinitha910 Can you merge the latest `master` into your branch? This should be sufficient:

```console
$ git checkout feature/pitcher_pouring
$ git fetch
$ git merge origin/master
```

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
+import numpy, openravepy, prpy, time
+from prpy import tsr
+from prpy.tsr import tsrlibrary
+from prpy.tsr.tsr import *
+
+@ActionMethod
+def MoveObject(robot, direction = [0, 0, 1], distance = 0.1):
```
mkoval: You should [*never* use a mutable object (a `list` in this case) as a default value in Python](http://stackoverflow.com/q/1132941/111426). Please replace this with `direction=None` and set the default inside the function body if `direction is None`.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+	"""
+	@param robot The robot performing the push grasp
+	@param direction The direction to move the object
+	@param distance The distance to move the object
+	"""
+    try:
+        manip = robot.GetActiveManipulator()
```
mkoval: You need to lock around `GetActiveManipulator()`.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+        manip.PlanToEndEffectorOffset(direction, distance, position_tolerance=0.1, execute=True)
+        return True
+    except Exception, e:
+        print 'MoveObject planning failed: ', str(e)
+        return False
```
mkoval: We shouldn't `print` in library functions - use a logger instead.

However, in this case, I think it would be better to avoid catching the exception all-together. I believe we typically indicate failure in an action by raising a `PlanningError` or `ActionError`; not returning `False`. @jeking04 can you confirm?

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+
+@ActionMethod
+def GraspHandle(robot, pitcher):
+    """
+    @param robot The robot performing the handle grasp
+    @param pitcher The pitcher who's handle will be grasped  
+    """
+    try:
+        # Get the pose of the pitcher
+        pitcher_in_world = pitcher.GetTransform()
```
mkoval: Missing lock.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Get the AABB of the pitcher
+        pitcher_aabb = pitcher.ComputeAABB()
```
mkoval: Missing lock.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+                                        [0., 0., 0., 1.]])
+
+        # we want a hand pose orthogonal to the direction of the spout
+        spout_direction = numpy.arctan2(spout_in_pitcher[0,1], spout_in_pitcher[0,0])
+        palm_direction = spout_direction - 0.5*numpy.pi
+
+        ee_in_pitcher = numpy.eye(4)
+        ee_in_pitcher[:3,:3] = numpy.array([[0, 0, 1],
+                                            [-1, 0, 0],
+                                            [0, -1, 0]])
```
mkoval: This `numpy.array()` call creates an integer array. I suggest adding at least one float (e.g. `0.`) to avoid any unpleasant surprises with clipping to integers.

In this specific case it's probably fine because you immediately store the `array` in a float `array` - but I'd still prefer if we didn't risk. it :smile:

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+        ee_in_pitcher[:3,:3] = numpy.dot(ee_in_pitcher[:3,:3],
+                                         openravepy.rotationMatrixFromAxisAngle([0, palm_direction, 0]))
+                                         
+        
+        offset = pitcher_aabb.extents()[0] + 0.28 # pitcher radius + ee_offset
+        ee_in_pitcher[:2,3] = -offset*ee_in_pitcher[:2,2]
+        ee_in_pitcher[2,3] = 0.45*pitcher_aabb.extents()[2]
+
+        ee_in_world = numpy.dot(pitcher_in_world, ee_in_pitcher)
+        manip = robot.GetActiveManipulator()
```
mkoval: Missing lock.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+        
+        with prpy.viz.RenderPoses([ee_in_world], manip.GetRobot().GetEnv()):
+            manip.PlanToEndEffectorPose(ee_in_world, execute=True)
+        manip.PlanToEndEffectorOffset(direction=manip.GetEndEffectorTransform()[:3,2], distance=0.1, execute=True)
```
mkoval: Missing lock around `GetEndEffectorTransform`.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+        manip.hand.CloseHand()
+        robot.Grab(pitcher)
+
+        return True
+    except Exception, e:
+        print 'GraspHandle planning failed: ', str(e)
+        return False
```
mkoval: See above. I suggest removing this `try`/`except` block.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+    @param pitcher The pitcher to perform the pouring  
+    """
+    # Now create a goal of tilting 
+    min_tilt_amount = 95. #degrees
+    max_tilt_amount = 100. #degrees
+    success = False
+    while not success and max_tilt_amount > 70.: 
+        try:
+            print 'Planning between %d and %d' % (min_tilt_amount, max_tilt_amount)
+            manip = robot.GetActiveManipulator()
```
mkoval: Missing lock around `GetActiveManipulator`?

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+
+@ActionMethod        
+def Pour(robot, pitcher):
+    """
+    @param robot The robot performing the pouring
+    @param pitcher The pitcher to perform the pouring  
+    """
+    # Now create a goal of tilting 
+    min_tilt_amount = 95. #degrees
+    max_tilt_amount = 100. #degrees
```
mkoval: It would be nice to make these arguments.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+                                             min_tilt = min_tilt_amount*numpy.pi/180., 
+                                             max_tilt = max_tilt_amount*numpy.pi/180.),
+                            smoothingitrs=100, execute=True)
+            success = True 
+        except Exception, e:
+            print 'Pour planning failed: ', str(e)
+            min_tilt_amount -= 5.
+            max_tilt_amount -= 5.
+            success = False
+    return success, min_tilt, max_tilt
```
mkoval: What's the reasoning behind doing this plan calls in a loop? If this is absolutely necessary, I'd prefer to do all the plan calls up-front, stitch the resulting trajectories together, and execute a the combined result.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+
+@ActionMethod
+def PushGraspCup(robot, cup, range):
+    """
+    @param robot The robot performing the push grasp 
+    @param cup The cup to push grasp
+    @param range The range for the yaw
+    """
+    try:
+        manip = robot.GetActiveManipulator()
```
mkoval: Missing lock.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+        manip.PlanToTSR(Cup_Grasp(robot, cup, manip, range), smoothingitrs=100, execute=True)
```
mkoval: Why are you passing `smoothingitrs` here? This should really only be used if you have trajectory-wide constraints.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+        manip.PlanToEndEffectorOffset(direction=manip.GetEndEffectorTransform()[:3,2], max_distance = 0.15, distance=0.01, execute=True)
```
mkoval: Missing lock around `GetEndEffectorTransform`.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+    except Exception, e:
+        print 'PushGraspCup planning failed: ', str(e)
+        return False
+
+def Cup_Grasp(robot, cup, manip, range, push_distance=0.0, **kw_args):
+	"""
+	@param robot The robot performing the push grasp
+	@param cup The cup to push grasp
+	@param manip The manipulator to perform the grasp with 
+       (if None active manipulator is used)
```
mkoval: Replace these tabs with spaces - they are not PEP-8 compliant and a recipe for difficult-to-debug errors.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+        manip.PlanToTSR(Cup_Grasp(robot, cup, manip, range), smoothingitrs=100, execute=True)
+        manip.PlanToEndEffectorOffset(direction=manip.GetEndEffectorTransform()[:3,2], max_distance = 0.15, distance=0.01, execute=True)
+        manip.hand.CloseHand()
+        manip.GetRobot().Grab(cup)
+        return True
+    except Exception, e:
+        print 'PushGraspCup planning failed: ', str(e)
+        return False
+
+def Cup_Grasp(robot, cup, manip, range, push_distance=0.0, **kw_args):
```
mkoval: Please rename this to match the naming convention used elsewhere in the file.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+	@param cup The cup to push grasp
+	@param manip The manipulator to perform the grasp with 
+       (if None active manipulator is used)
+    @param range The range for the yaw
+    @param push_distance The distance to push before grasping
+	"""
+    if manip is None:
+        manip_idx = robot.GetActiveManipulatorIndex()
+    else:
+        manip_idx = manip.GetRobot().GetActiveManipulatorIndex()
```
mkoval: Missing lock.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+
+    T0_w = cup.GetTransform()
+    T0_w[:3, :3] = robot.GetTransform()[:3, :3]
```
mkoval: Missing locks around `GetTransform`.

src/herbpy/action/actions.py
```diff
@@ -0,0 +1,145 @@
...
+                        [0., 0., 0., 1.]])
+
+    Bw = numpy.zeros((6,2))
+    Bw[2,:] = [0.03, 0.04]  # Allow a little vertical movement
+    Bw[5,:] = range
+    
+    grasp_tsr = prpy.tsr.TSR(T0_w = T0_w, Tw_e = Tw_e, Bw = Bw, manip = manip_idx)
+    grasp_chain = prpy.tsr.TSRChain(sample_start=False, sample_goal = True, 
+                                    constrain=False, TSR = grasp_tsr)
+    return [grasp_chain]
```
mkoval: Please add a newline to the end of the file.

src/herbpy/action/actions.py
mkoval: Also, shouldn't this be handled by the `TSRLibrary`?

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+from herbpy.tsr.pitcher import *
+
+def PourWater(robot, cup, pitcher):
+	"""
+	@param robot The robot performing the pouring
+	@param cup The cup to pour into
+	@param pitcher The pitcher to pour
+	"""	
+
+	manip = robot.left_arm
```
mkoval: Tabs.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+    manip_cup = robot.right_arm
```
mkoval: It would be nice to make this an argument. Or does it only work on the right arm?

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+
+	# Grasp the handle
+    success = actions.GraspHandle(robot, pitcher)
+
+    #Lift Pitcher
+    if success:
+        success = actions.MoveObject(robot, direction=[0, 0, 1], distance=0.06)
+    
+    manip_pose = robot.GetActiveManipulator().GetEndEffectorTransform()
```
mkoval: Missing lock.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+    success = actions.GraspHandle(robot, pitcher)
+
+    #Lift Pitcher
+    if success:
+        success = actions.MoveObject(robot, direction=[0, 0, 1], distance=0.06)
+    
+    manip_pose = robot.GetActiveManipulator().GetEndEffectorTransform()   
+    
+    manip_cup.SetActive()
+    robot.SetActiveManipulator(manip_cup)
```
mkoval: Missing lock.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+    
+    #Grasp cup
+    success_cup = actions.PushGraspCup(robot, cup)
+
+    pitcher_start = pitcher.GetTransform()
+    pitcher_aabb = pitcher.ComputeAABB()
```
mkoval: Missing lock.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+    if success_cup:
+        success_cup = actions.MoveObject(robot, direction=[0, 0, 1], distance=0.01)
+
+    manip.SetActive()
+    robot.SetActiveManipulator(manip)
+
+    if success_cup:
+        p_tsr, min_tilt, max_tilt = robot.tsrlibrary(pitcher, 'pour')
+        traj = manip.PlanToTSR(p_tsr, execute=False)
+        cspec = traj.GetConfigurationSpecification()
```
mkoval: I think you want to use `GetActiveConfigurationSpecification` here. You may also need to set `manip.GetArmIndices()` as active earlier in this function to make this work as expected.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+        wpt = traj.GetWaypoint(traj.GetNumWaypoints() - 1)
+        config = cspec.ExtractJointValues(wpt, robot, manip.GetArmIndices())
```
mkoval: Lock around `GetArmIndices`.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+        success_cup = actions.MoveObject(robot, direction=[1, 0, 0], distance=cup_aabb.extents()[0]/2)
+
+    if success_cup:
+        success_cup = actions.MoveObject(robot, direction=[0, -1, 0], distance=cup_aabb.extents()[1])
+
+    if success_cup:
+        success_cup = actions.MoveObject(robot, direction=[0, 0, -1], distance=0.009)       
+
+    manip.SetActive()
+    robot.SetActiveManipulator(manip)
```
mkoval: Missing locks. This second line is redundant - it does the same thing as the first.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # Pour
+    # slow the arm down
+    v = numpy.array([0.75, 0.75, 2., 2., 2.5, 2.5, 2.5])
+    robot.GetActiveManipulator().SetVelocityLimits(0.5*v, 0.3)
```
mkoval: Missing lock.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+    
+    #Gets position of the pitcher and pours
+    if success and success_cup:
+        traj = robot.PostProcessPath(traj)
+        robot.ExecuteTrajectory(traj)
+
+    time.sleep(2)
+    
+    #Slows down the speed of the arm
+    robot.GetActiveManipulator().SetVelocityLimits(v, 0.3)
```
mkoval: Missing lock.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+    pos = pitcher_aabb.pos()[2]
+    move_down = pos - table_height - pitcher_half
+
+    if success:
+        success = actions.MoveObject(robot, direction=[0,0,-1], distance=move_down-0.05)
+    
+    #Open the hand and release the pitcher
+    if success:
+        manip.hand.OpenHand()
+        robot.Release(pitcher)
```
mkoval: Add a new line to the end of the file.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+    robot.GetActiveManipulator().SetVelocityLimits(v, 0.3)
+
+    # Tilts pitcher back to original position
+    if success:
+        robot.ExecuteTrajectory(openravepy.planningutils.ReverseTrajectory(traj))
+    
+    if success:
+        success = actions.MoveObject(robot, direction=-1.*manip_pose[:3,0], distance=0.1)
+    
+    pitcher_aabb = pitcher.ComputeAABB()
```
mkoval: Missing lock.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
...
+            manip.SetDOFValues(config)
+            desired_cup_in_world_x = pitcher_start[0,3] + (2*pitcher_aabb.extents()[0])
+            desired_cup_in_world_y = pitcher_start[1,3] - (2*pitcher_aabb.extents()[1])
+            cup_in_world_x = cup_pose[0, 3]
+            cup_in_world_y = cup_pose[1, 3]
+            move_x = desired_cup_in_world_x - cup_in_world_x
+            move_y = desired_cup_in_world_y - cup_in_world_y
+            move_cup = math.sqrt(math.pow(move_x, 2) + math.pow(move_y, 2))
+        manip_cup.SetActive()
+        robot.SetActiveManipulator(manip_cup)
```
mkoval: Redundant. `SetActive` already does this.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,104 @@
+import openravepy, os, time, prpy, actions, math
+from prpy.action import ActionMethod
+from herbpy.tsr.cup import *
+from herbpy.tsr.pitcher import *
+
+def PourWater(robot, cup, pitcher):
```
mkoval: This function changes the robot's active manipulator, active DOFs, and velocity limits. You should wrap the relevant portions of the function in a robot state saver to restore them to the defaults before returning.

mkoval:
A few general comments:

1. Actions should raise an exception, typically a `PlanningError` or an `ActionError`, if they fail. I don't think there is any expectation that the user will check the return code for success.
2. Consider adding a `manip` argument to most of these functions. We typically set this to `GetActiveManipulator` by default.
3. Calls to `robot.planner` use the active DOFs and active manipulator, so you can safely call `robot.PlanToXXX` instead of `manip.PlanToXXX` if those attributes are set correctly. This eliminates an extra environment clone and can actually lead to a significant performance improvement.
4. Please wrap any changes to the robot's state, e.g. active DOFs or limits, in a `RobotStateSaver` that restores those attributes to their original values.
5. Be sure to lock the environment around any OpenRAVE calls that are not explicitly listed as thread safe (e.g. `GetTransform`, `GetDOFValues`, etc).

I left some specific comments above. Tagging @jeking04 to make sure I didn't give you any bad information. :smile:

vinitha910:
I will make the necessary changes let you know when I'm done. Thanks

vinitha910:
I believe everything is fixed now. @mkoval  Can you take a look at it? Thanks

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,72 @@
...
+def MoveCupAndPour(robot, table, manip_pitcher, cup, pitcher):
+    """
+    @param robot The robot performing the pouring
+    @param table The table the objects are on
+    @param manip_pitcher The arm that is manipulating the pitcher
+    @param cup The cup to pour into
+    @param pitcher The pitcher to pour
+    """	
+
+    manip_cup = robot.right_arm
```
mkoval: Please take the `manipulator` as an argument. I would also prefer if it defaulted to `GetActiveManipulator()`, instead of the `right_arm`.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,72 @@
...
+        cup_aabb = cup.ComputeAABB()
+        manip_cup_pose = manip_cup.GetEndEffectorTransform()
+        manip_pitcher_pose = manip_pitcher.GetEndEffectorTransform()
+
+    p_tsr, min_tilt, max_tilt = robot.tsrlibrary(pitcher, 'pour')
+    traj = manip_pitcher.PlanToTSR(p_tsr, execute=False)
+    cspec = traj.GetConfigurationSpecification()
+    wpt = traj.GetWaypoint(traj.GetNumWaypoints() - 1)
+    config = cspec.ExtractJointValues(wpt, robot, manip_pitcher.GetArmIndices())
+    with env, robot:
```
mkoval: Please create an explicit `RobotStateSaver` with the flags you care about, instead of using the omnibus `with robot` form. In this case, I believe you only need `LinkTransformations`.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,72 @@
...
+        manip_pitcher.SetDOFValues(config)
+        desired_cup_in_world_x = pitcher_start[0,3] + (2*pitcher_aabb.extents()[0])
+        desired_cup_in_world_y = pitcher_start[1,3] - (2*pitcher_aabb.extents()[1])
+        cup_in_world_x = cup_pose[0, 3]
+        cup_in_world_y = cup_pose[1, 3]
+        move_x = desired_cup_in_world_x - cup_in_world_x + cup_aabb.extents()[0]/2
+        move_y = desired_cup_in_world_y - cup_in_world_y - cup_aabb.extents()[1]
+        move_cup = math.sqrt(math.pow(move_x, 2) + math.pow(move_y, 2))
```
mkoval: I suggest replacing this with `numpy.linalg.norm`.

src/herbpy/action/pouring.py
```diff
@@ -0,0 +1,72 @@
...
+        
+        #Gets position of the pitcher and pours
+        traj = robot.PostProcessPath(traj)
+        robot.ExecuteTrajectory(traj)
+        time.sleep(1) #To let the water pour from pitcher into cup
+    
+        #Slows down the speed of the arm
+        manip_pitcher.SetVelocityLimits(v, 0.3)
+    finally:
+        manip_pitcher.SetVelocityLimits(v_original, 0.3)
```
mkoval: Please use a `RobotStateSaver` with the appropriate options`, instead of doing the restoration yourself. Note that there is a bug here - you're assuming that the `min_accel_time` is 0.3.

src/herbpy/tsr/pitcher.py
```diff
@@ -1,9 +1,10 @@
-import numpy
+import numpy, openravepy
+import prpy.tsr
 from prpy.tsr.tsrlibrary import TSRFactory
 from prpy.tsr.tsr import TSR, TSRChain
-
-@TSRFactory('herb', 'rubbermaid_ice_guard_pitcher', 'grasp')
-def pitcher_grasp(robot, pitcher, manip=None):
+      
+@prpy.tsr.tsrlibrary.TSRFactory('herb', 'rubbermaid_ice_guard_pitcher', 'push_grasp')
```
mkoval: You already `import`ed `TSRFactory`, so this full-qualification is not necessary.

mkoval:
Looks good other than a few nit-picks. @vinitha910 thanks for making those changes!

vinitha910:
@mkoval I made the last few changes, can you take another look at it? Thanks!

mkoval:
@vinitha910 Thanks for making those changes!

PR #80, opened 2016-02-22
Added yaw_range parameter to glass grasps

vinitha910:
I added yaw_range to the glass grasping TSRs and the PushGrasp function. If no yaw_range is given the default is [-numpy.pi, numpy.pi]. I also, added **kw_args to pitcher.py.


mharding01:
Looks good, but you might need to elaborate on the form of yaw_range input (i.e. 'list' with more on what contents look like). You have None being equivalent to [-pi, pi], but would this have the same effect if you input [-pi, pi]? - just for consistency.

vinitha910:
@mharding01 I'm not sure what you mean by elaborating on the form of yaw_range input. None is equivalent to [-pi,pi] and this would be equivalent to inputting [-pi,pi]. The None is there if the user decides that they don't want to restrict the yaw_range.

PR #81, opened 2016-02-25
Adding named preshapes for the hands.

jeking04:
This is an initial attempt at providing named preshapes for hands that follow the same model as named configurations for the robot.

### DIFF
--- a/config/barrett_preshapes.yaml
+++ b/config/barrett_preshapes.yaml
@@ -1,0 +1,11 @@
+configurations:
+    closed:
+        hand: [2.443, 2.443, 2.443, 0.0]
+    open:
+        hand: [0., 0., 0., 0.]
+    bowl_grasp:
+        hand: [0.977, 0.977, 0.977, 0.0]
+    glass_grasp:
+        hand: [0.488, 0.488, 0.488, 0.0]
+    tray_handle_grasp:
+        hand: [0.977, 0.977, 0.977, 0.0]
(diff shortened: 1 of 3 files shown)


mkoval:
I would prefer to put all of these in the `robot.configurations` object.

One motivation behind creating the `ConfigurationLibrary` class, instead of simply using a `dict`, was to encapsulate all of unpleasant logic for handling DOF indices in one place. For example:

- querying a subset of the DOF indices stored in the library
- mapping between the input DOF indices and the order stored in the library
- handling duplicate DOF indices
- checking for invalid DOF indices

Therefore, we should be able to support hand and arm configurations in the same library. If this is not the case, e.g. some of this functionality is not implemented, then I would prefer to fix it rather than proliferate more instances of `ConfigurationLibrary`.

We can still add a helper function on `robot.hand` that queries the `ConfigurationLibrary` with the right DOF indices.

jeking04:
I agree that it would be nice to put everything in one ```robot.configurations``` object. Like you mentioned, I think we would have to add something to ```prpy.base.EndEffector``` that allowed us to make a call such as ```hand.GetNamedPreshape('grasp')```.

Here is where I got stuck:
I can update the barret config to specify ```right_hand``` and ```left_hand``` instead of just ```hand``` and then load the barret config into ```robot.configurations```.

But imagine, hypothetically, that we wanted to be able to swap out different types of hands on the robot, i.e. replace the Barrett with a Robotiq. My thought was that we would create separate config files for each hand and load the correct one based on hand type. But if the hands don't match, i.e. left hand Barrett, right hand Robotiq, then loading the Robotiq configurations will overwrite the ```left_hand``` Barrett configs.

Maybe we add an optional parameter to ```load_yaml``` in ```ConfigurationsLibrary``` that lets us specify to only load a particular group? Other ideas?

mkoval:
That's a very good point. I also can't think of a better solution than what you've already implemented - putting the `ConfigurationLibrary` inside the hand.

PR #82, opened 2016-03-21
Added HRC task support with skeleton tracking integration

stepelle:
I added some functions through which it is possible to (1) use the kinect mounted on Herb to track people and to add them the openrave env (2) execute the human-robot cooperation task in which herb stamps some boxes and the human ties a ribbon on them.


src/herbpy/herbrobot.py
```diff
@@ -271,7 +270,72 @@ def SetStiffness(self, stiffness):
...
+                logger.info('Humans_tracking')
+            if hrc==True:
+                from hrc import assistance                 
+                humans_hrc = []
+                logger.info('HRC')
+                
+        try:
+            rospy.init_node(node_name, anonymous=True)
+        except rospy.exceptions.ROSException:
+            pass
```
mkoval: We can only call `init_node` once per process, so we always do so in the main script where we call `herbpy.initialize` to prevent multiple modules from conflicting with each other. Please change this function to throw an exception if a ROS node is not already running, instead of starting one yourself.

src/herbpy/herbrobot.py
```diff
@@ -271,7 +270,72 @@ def SetStiffness(self, stiffness):
         self.head.SetStiffness(stiffness)
         self.left_arm.SetStiffness(stiffness)
         self.right_arm.SetStiffness(stiffness)
-
+        
+    def DetectHuman(self, env, orhuman=2, load_hum=True, 
+                    hrc=False, continuous=True,      
+                    herb_sim=True, enable_legs=True,             
+                    segway_sim=True, action='stamp'):
```
mkoval: Please an a docstring for each argument. Most functions in PrPy are documented properly, so you can look there for examples.

src/herbpy/herbrobot.py
```diff
@@ -271,7 +270,72 @@ def SetStiffness(self, stiffness):
...
+        """Use the kinbody detector to detect objects and add
+        them to the environment
+        """
+        import rospy
+        from tf import TransformListener, transformations
+
+        node_name = 'detecthuman'  
+
+        if orhuman==0:
```
mkoval: What is the meaning of the `orhuman` argument? It looks like an enum that controls the method used to track the skeleton. If that is the case, I suggest using a meaningful string for each choice instead of a number.

src/herbpy/herbrobot.py
```diff
@@ -271,7 +270,72 @@ def SetStiffness(self, stiffness):
...
+        elif orhuman==2:
+            if load_hum==True:
+                #kin 2 - load the human from openrave
+                import humanpy.hum_kin2 as sk
+                humans = []
+                logger.info('Humans_tracking')
+            if hrc==True:
+                from hrc import assistance                 
+                humans_hrc = []
+                logger.info('HRC')
```
mkoval: What if `orhuman` is not `0`, `1`, or `2`? If I understand this correctly, that is an error and you should throw an exception.

src/herbpy/herbrobot.py
```diff
@@ -271,7 +270,72 @@ def SetStiffness(self, stiffness):
...
+        try:
+            rospy.init_node(node_name, anonymous=True)
+        except rospy.exceptions.ROSException:
+            pass
+           
+        try:  
+            tf = TransformListener()  
+            humanadded = False
+
+            while not rospy.is_shutdown():
```
mkoval: This is an infinite loop. How is this intended to work in process? Do you run this function in another thread?

src/herbpy/tsr/box.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @param robot The robot to perform the stamp
+    @param box The box to stamp
+    @param manip The manipulator to stamp 
+    '''
+
+    if manip is None:
+        manip_idx = robot.GetActiveManipulatorIndex()
+    else:
+        with manip.GetRobot():
```
mkoval: Please use `CreateRobotStateSaver` to create a state saver that *only* restores the active manipulator instead of relying on the default behavior of `with robot`.

src/herbpy/tsr/box.py
```diff
@@ -0,0 +1,45 @@
...
+            manip.SetActive()
+            manip_idx = manip.GetRobot().GetActiveManipulatorIndex()
+
+    T0_w = box.GetTransform()
```
mkoval: Missing lock.

mkoval:
In addition to those comments above, please revert the permissions on the `src/herbpy/tsr` directory. It looks like you may have accidentally made all contents of the directory executable.

src/herbpy/tsr/box.py
stepelle: In relation to this question and the question in which you require to lock, do you mean this?
 
    with robot.GetEnv():
        with robot.CreateRobotStateSaver():
            if manip is None:
                manip_idx = robot.GetActiveManipulatorIndex()
                manip = robot.GetActiveManipulator()
            else:
                manip.SetActive()
                manip_idx = manip.GetRobot().GetActiveManipulatorIndex()

        T0_w = box.GetTransform()

src/herbpy/herbrobot.py
stepelle: Done. The exception is thrown by the following try/except.

src/herbpy/herbrobot.py
stepelle: Done

src/herbpy/herbrobot.py
stepelle: I explained the meaning of orhuman and I changed the enum into strings.

src/herbpy/herbrobot.py
stepelle: Exception added

Stefanos19:
Could we please merge this? This branch is required for Stefania's `humanpy` code to run.

src/herbpy/herbrobot.py
```diff
@@ -271,7 +270,93 @@ def SetStiffness(self, stiffness):
...
+        @param action if hrc=True, the kind of action the herb is going to perform 
+                      between stamping ('stamp') and grasping ('grasp')
+        """
+        
+        
+        import rospy
+        from tf import TransformListener, transformations
+
+        if orhuman=='kin1_skel': 
+            if load_hum==True:
```
mkoval: Use `if load_hum`. Comparing against `True` and `False` is an anti-pattern.

src/herbpy/herbrobot.py
```diff
@@ -271,7 +270,93 @@ def SetStiffness(self, stiffness):
         self.head.SetStiffness(stiffness)
         self.left_arm.SetStiffness(stiffness)
         self.right_arm.SetStiffness(stiffness)
-
+        
+    def DetectHuman(self, env, orhuman='kin2_or', load_hum=True,                           
+                    enable_legs=True, segway_sim=True, 
+                    base_frame='/map' , kin_frame='/head/skel_depth_frame2',
+                    hrc=False, herb_sim=True, 
+                    continuous=True, action='stamp'):
```
mkoval: I don't think this should be a function HerbPy. It enters an infinite `while not rospy.is_shutdown()` loop and will never return in normal operation. I suggest moving this to a separate script file, much like `console.py`, that we can run from the command line for visualization purposes.

src/herbpy/herbrobot.py
```diff
@@ -271,7 +270,93 @@ def SetStiffness(self, stiffness):
...
+                #kin 2 - load the human from openrave
+                import humanpy.humantracking_kinect2 as sk
+                humans = []
+                logger.info('Humans_tracking')
+            if hrc==True:
+                from hrc import assistance                 
+                humans_hrc = []
+                logger.info('HRC')
+        else:
+            raise Exception('orhuman param is not correct')
```
mkoval: This giant chain of `if` statements concerns me. It seems like most of the options are meaningless unless `load_hum` is `True`. Is there any way we can simplify this logic and/or split it into separate functions? Can someone explain what is going on here?

src/herbpy/tsr/box.py
```diff
@@ -0,0 +1,47 @@
...
+    '''
+    This creates a TSR for stamping a box. 
+    It is assumed that when called, the robot is grasping a stamp
+
+    @param robot The robot to perform the stamp
+    @param box The box to stamp
+    @param manip The manipulator to stamp 
+    '''
+    
+    with robot.GetEnv():
```
mkoval: This lock is not necessary. The environment is guaranteed to be locked around a `tsrlibrary` call.

@jeking04, can you confirm?

src/herbpy/tsr/box.py
```diff
@@ -0,0 +1,47 @@
...
+        with robot.CreateRobotStateSaver():
```
mkoval: Pass the appropriate options into `CreateRobotStateSaver`. In this case, you need to save the active manipulator and active DOFs.

src/herbpy/tsr/box.py
```diff
@@ -0,0 +1,47 @@
...
+    palm_to_box_center = .045
+    adjustment = -0.09
+    total_offset = ee_to_palm + palm_to_box_center + adjustment
+    Tw_e = numpy.array([[ 0., 0., 1., -total_offset], 
+                        [1., 0., 0., 0.], 
+                        [0., 1., 0., 0.38], # box height 0.2  0.35
+                        [0., 0., 0., 1.]])
+
+    Bw = numpy.zeros((6,2))
+    #Bw[2,:] = [0.0, 0.00]  # Allow a little vertical movement
```
mkoval: Remove dead code.

mkoval:
@Stefanos19 I finally reviewed this. Sorry for the delay. 😓 

Some high-level comments:

- [ ] Merge `master` into this branch.
- [ ] Remove `cmovetraj.txt`
- [ ] Fix file permissions on `__init__.py`, `herb.py`, `herbbase.py`, `herbpantilt.py`, and `herbrobot.py`.
- [ ] Move `DetectHuman` into a standalone Python script (or present a compelling argument otherwise).

I also left a handful of more specific comments inline.

stepelle:
@mkoval: I made all the changes and I merged the master in the branch. I moved the DetectHuman function to another package (humanpy). However, I cannot test the code.... @Stefanos19: can you do it?

src/herbpy/tsr/box.py
```diff
@@ -0,0 +1,45 @@
...
+    @param manip The manipulator to stamp 
+    '''
+  
+    with robot.CreateRobotStateSaver(robot.GetActiveManipulatorIndex() | robot.ActiveDOF()):
+      if manip is None:
+        manip_idx = robot.GetActiveManipulatorIndex()
+        manip = robot.GetActiveManipulator()
+      else:
+        manip.SetActive()
+        manip_idx = manip.GetRobot().GetActiveManipulatorIndex()
```
mkoval: Replace this with `prpy.util.GetManipulatorIndex`.

PR #89, opened 2016-05-24
Feature/generic tsrs

jeking04:
This PR removes the bowl, plate and tray TSRs that are fairly specific to table clearing. It also updates several TSRs to use generic TSRs added as part of prpy PR 309:

https://github.com/personalrobotics/prpy/pull/309


src/herbpy/tsr/fuze.py
```diff
@@ -31,29 +31,46 @@ def _fuze_grasp(robot, fuze, push_distance = 0.0, manip = None, **kw_args):
...
-    grasp_chain = TSRChain(sample_start=False, sample_goal = True, 
-                           constrain=False, TSR = grasp_tsr)
-
-    return [grasp_chain]
+    @param robot The robot grasping the fuze
+    @param fuze The grasped object
+    @param pose_tsr_chain The tsr chain for sampling placement poses for the fuze
+    @param manip The manipulator grasping the object, if None the active
+       manipulator of the robot is used
+    '''
```
mkoval: `'''` to `"""`.

src/herbpy/tsr/fuze.py
```diff
@@ -31,29 +31,46 @@ def _fuze_grasp(robot, fuze, push_distance = 0.0, manip = None, **kw_args):
...
+    configuration for transport.
 
+    @param robot The robot grasping the fuze
+    @param fuze The grasped object
+    @param manip the manipulator grasping the object, if None the active manipulator 
+       of the robot is used
+    @param roll_epsilon The amount to let the object roll during transport (object frame)
+    @param pitch_epsilon The amount to let the object pitch during transport (object frame)
+    @param yaw_epsilon The amount to let the object yaw during transport (object frame)
+    '''
```
mkoval: `'''` to `"""`.

src/herbpy/tsr/table.py
```diff
@@ -37,3 +37,46 @@ def point_on(robot, table, manip=None):
...
+
+@TSRFactory('herb', 'table', 'table_edge')
+def table_edge(robot, table, **kwargs):
+    '''
+    This creates a TSR that allows you to sample poses from either
+    long edge of the table.
+    
+    @param robot The robot (unused)
+    @param table The table
+    '''
```
mkoval: What is the convention of these frames? z-up? What about the x and y-axes?

jeking04:
I believe I have now addressed all the comments on this PR. The checks are failing because of the dependency on prpy #309. But I have verified tests pass locally. Once we merge the prpy PR we can restart the travis checks and then this PR should be good to go.

mkoval:
This pull request now contains #95. Is that intentional?

Otherwise, I'm happy with this pull request. I will merge it once @ClintLiddick signs on #95 and we merge that pull request.

jeking04:
It is intentional. We needed the changes in both PRs for some testing so we pulled #95 into this one. We should merge #95 first.

PR #98, opened 2016-06-22
Allow passing object sampling tsr to place function

Mehar-Singh:


src/herbpy/herbrobot.py
```diff
@@ -225,7 +225,7 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
                      TSRPlanner(delegate_planner=actual_planner),
                      self.cbirrt_planner),
             # Special purpose meta-planner.
-            NamedPlanner(delegate_planner=actual_planner),
+            NamedPlanner(delegate_planner=Sequence(actual_planner, self.cbirrt_planner)),
```
mkoval: What is the purpose of this change?

src/herbpy/action/grasping.py
```diff
@@ -189,14 +189,15 @@ def Lift(robot, obj, distance=0.05, manip=None, render=True, **kw_args):
...
-def Place(robot, obj, on_obj, manip=None, render=True, **kw_args):
+def Place(robot, obj, on_obj, obj_top_tsr = None, manip=None, render=True, **kw_args):
     """
     Place an object onto another object
     This assumes the 'point_on' tsr is defined for the on_obj and
     the 'place' tsr is defined for obj
     @param robot The robot performing the push grasp
     @param obj The object to place
     @param on_obj The object to place obj on
+    @param obj_top_tsr TSR describes that can sample the poses
```
mkoval: Fix grammar in this comment.

src/herbpy/herbrobot.py
jeking04: This allows calls to ```PlanToNamedConfiguration``` to use a randomized planner (in this case ```CBiRRT```) when snap planner and optimizers have failed. We are seeing that when planning out of tight quarters near objects using CHOMP the resulting trajectory fails to get pushed out of collision.  I think this change is important, especially if a user has only CHOMP installed and not trajopt, but we will pull it out and put it into a separate PR.

mkoval: That is reasonable. I am fine with leaving it in this pull request.

PR #99, opened 2016-07-01
removed glass and table tsrs

athackst:
It looks like bowl and plate were already removed but these weren't.


jeking04:
I actually left these on purpose. These TSRs can be used independently of table clearing. For example, we use the glass ```grasp``` and table ```point_on``` TSRs in our Grasping and Placing actions here:

https://github.com/personalrobotics/herbpy/blob/master/src/herbpy/action/grasping.py

For "quick hit" demos and testing, i.e. we can run:
```
cmd> rosrun herbpy console.py
ipython> robot.DetectObjects()
ipython> glass = ...
ipython> table = ...
ipython> robot.Grasp(glass)
ipython> robot.Place(glass, table)
```

I removed the plate and bowl TSRs from herbpy because they were strongly tied to table clearing. For example, the ```grasp``` TSR for the bowl didn't actually move the hand to a location where you can just close the fingers and achieve a grasp. Instead, it relied on the sequence of actions specified in table clearing: move to the ```grasp``` TSR, move down to touch the table, close the fingers.  Neither the bowl nor the plate are conducive to these "quick hit" demos. They require longer sequences of actions that are better achieved through HGPC.

So my preference would be to keep these TSRs in herbpy.

athackst:
Ok, gotcha.

athackst:
Is this just because you don't want to rely on external packages?  I was hoping to move from robot-specific tsrs to something more general and I'm a little wary of having multiple definitions of the same object out there.

jeking04:
Its a good point. I am okay with relying on external packages if they are public. I don't want ```herbpy``` to depend on external packages that are only accessible to people on the Toyota project.

jeking04:
Sorry just hit the wrong button.

athackst:
Ok, the way I see it, there are a couple of options here:

1. We don't make any changes and we have to support 2 tsrs for the same object
2. We remove the include in __init__ which would encourage usage of the general tsr, but not prohibit the use of the herb tsr.  We would still have to support 2 tsrs for the same object, but hopefully the general one is more up-to-date
3. "quick hit" demos and testing are done through a separate package that depends on the external library.  However that will be limited to our project.
4. I get the external project open sourced.  This is not likely to happen soon.
5. herbpy shared tsrs are auto-genned (compile? script?) from the generic tsrs and thou shalt not manually edit them in herbpy.

thoughts?  I'm leaning toward the last option at the moment.

jeking04:
I think 3 and 4 are out. For 3, people in the lab outside of the project depend on this functionality, so we need to keep it.

I am trying to decide between 1, 2, and 5. To help me understand which of these options is best, can you point me to where "pinch", "cylinder" and "hook" are resolved? Basically, what I want to know is are we really maintaining support for two TSRs, or two sets of herb-specific parameters to use on those TSRs.

Also, I am having trouble understanding/finding where the table TSRs moved to and how those fit into this framework. I think maybe I am missing a branch/fork of prpy?

athackst:
I'm not sure I understand your question, but 
gripper_tsr.combine_tsr is where the grasps get combined with the object.

By keeping both, if you want to add something to glass (like combining "grasp thumb up" and "grasp thumb down" into one grasp tsr chain), and you only add it to herbpy's glass, only herb will be able to use it, or you will have to _also_ add it to the generic glass, which is more work and prone to divergence between robots.

I think I moved table-specific tsrs into generic so we could use them for both "wooden table" and "conference table", since they deal with object extents that can be calculated for any mesh, but I can add a table file if that's better.

For prpy you will need my branch,  I've made a pull request.

jeking04:
What I was missing was what "generic" grasp meant. But I think I get it now, its defined here:

https://bitbucket.org/toyotarobotics/tsr_gripper_library/src/ef9b068f6073a91fb7b1a53dd3875057a53fe904/grippers/barrett_hand/grasps.yaml?at=master&fileviewer=file-view-default

I think this is basically the yaml version of the generic cylinder TSR defined here:

https://github.com/personalrobotics/prpy/blob/master/src/prpy/tsr/generic.py#L20

Which is what the HERB tsr is calling, with bindings on the parameters:

https://github.com/personalrobotics/herbpy/blob/master/src/herbpy/tsr/glass.py#L47

Which mimics your yaml HERB specific grasp data:

https://bitbucket.org/toyotarobotics/tsr_gripper_library/src/ef9b068f6073a91fb7b1a53dd3875057a53fe904/robots/herb/left_gripper.yaml?at=master&fileviewer=file-view-default

I think the best thing to do is go with option 2 above. I don't mind option 5 except I feel a bit uncomfortable with not allowing anyone in the lab outside the toyota project to edit these.  Ideally, we open source the tsr gripper library  and switch to it.  

I do think we have to be a little careful calling the grasps in the ```grasps.yaml``` generic. They are robot agnostic I agree. But they are specific to both gripper and object, i.e. the cylinder grasp is defined to grasp the object at 0.08 meters above it center, which is likely specific to the glass.

My only problem with the table TSRs is that I am pretty sure the wooden table and the conference table have different coordinate frames: The conference table has y-axis pointing through the top, the wooden table has z (see attached). So we actually need to define the TSRs differently for them.
table_frames

athackst:
I thought about this over the weekend, and I think that it might be better to just move
```
import herbpy.tsr
```
from herbrobot.py to console.py

That way it's not automatically included with table clearing, but it is automatically included for your easy scripts.  It's also a little more sane to add the entire folder rather than just some of them.  If we want to pull in other objects to the tsr_gripper_library or update them we can do it separately.  

--The grasps in grasp.yaml are intended to be _gripper specific_ and not gripper **and** object specific. So they are "general" in that sense.  For example, the same definition for "cylindar" grasp should be able to be used on a wide array of objects (glass, 2x4, water bottle, etc.)

--For the table.  I believe ComputeAABB() gives the results in world frame, unless I misunderstood it?  So it shouldn't matter what the frame of the table is, so long as the table is "up right"

jeking04:
1. Sounds good. I'll make that change and put in a PR.

2. I think I might still be misunderstanding something here. Doesn't the 0.08 component here imply something about the shape of the object:
https://bitbucket.org/toyotarobotics/tsr_gripper_library/src/ef9b068f6073a91fb7b1a53dd3875057a53fe904/grippers/barrett_hand/grasps.yaml?at=master&fileviewer=file-view-default#grasps.yaml-22

In particular, if the object is less than 8 cm tall, or the reference frame is defined less than 8 cm from the top of the object, the grasp will always miss? Similarly with hook:

https://bitbucket.org/toyotarobotics/tsr_gripper_library/src/ef9b068f6073a91fb7b1a53dd3875057a53fe904/grippers/barrett_hand/grasps.yaml?at=master&fileviewer=file-view-default#grasps.yaml-39

It seems the 0.11 and 0.17 are encoding information about the object this grasp was designed for?

It feels like these generic TSRs need at some point to know some information about the object they are being applied to in order to instantiate them, i.e. the cylinder TSR needs to know where vertically on the cylinder you want the grasp to occur.

3. I see. You are definitely correct in that ```ComputeAABB``` returns in world frame. And I missed this line: https://bitbucket.org/toyotarobotics/tsr_gripper_library/src/ef9b068f6073a91fb7b1a53dd3875057a53fe904/src/tsr_gripper_library/tsr/generic.py?at=master&fileviewer=file-view-default#generic.py-16

which ignores the orientation component of the table when setting ```T0_w```.  This does correctly allow this code to work regardless of the table's frame. However, this also binds the TSR to the world frame. Meaning all samples will be along the world y component. Rotating the table by 90 degrees will make the TSR return along a different edge. Rotation by 45 degrees will make the TSR return something not along any edge.  I think we need to treat the table like a robot and perform the same setup that you are for grippers. Basically, we need an extra transform that tells us how to go from the frame of the table to some task frame (the equivalent of the grasp frame).

athackst:
--on 2.)  No .8 has nothing to do with the object, it's the offset from the bh_base_link to a point on the palm (or maybe just in front of the palm).  Similar with hook; it's the offset from bh_base_link to a point at the tip of the fingers, which is offset in both y and z from bh_base_link.

Of course these are grasp definitions that are particular to a specific gripper.  In order to pick up an object you will also need an object specific TSR.  These two get combined together to create a gripper/object TSR which is actionable.  The _object TSR_ encodes orientation and location of an graspable frame w.r.t. the objection (and the assumption that the frame is defined in a very specific way) and the _grasp definition_ encodes placement in hand.  So if instead of grabbing the glass, you wanted to poke the glass, you can use the same glass TSR definition (frames oriented around the outside of the glass cylindar) and combine it with a "poke" definition which defines the offset from ee to fingertip, and now you have a TSR for touching the glass with a finger.

On the table-- I see, yes, I was assuming both "is upright" and "is perpendicular" which is not a good assumption.  The table will need it's own object-specific TSRs.  I vote for the inherited version you created.

jeking04:
I get it now! Finally, I know.

Okay, I think removing the import is a fine thing to do for now. 

Here is another option: 

1. Move ```combine_tsr``` into a ```prpy``` utility function 

2. Move the TSRs defined here:

https://bitbucket.org/toyotarobotics/tsr_gripper_library/src/ef9b068f6073a91fb7b1a53dd3875057a53fe904/src/tsr_gripper_library/tsr/?at=master

into ```table_clearing``` or ```prpy```. 

3. Still remove the import from herbpy. But update the herbpy TSRs to just instantiate those reference frames in python and still use the generic object TSRs.  Then, rather than fully duplicating the TSRs, we are really just duplicating the offsets defined in ```left_gripper.yaml``` etc.

My guess is that step 1 is a no go, so we can just go with the removing the ```import``` call for now.

athackst:
I think moving the import call is the best for now.

PR #103, opened 2016-08-05
Enabled baked collision checking by default

mkoval:
This should be merged after #102 and:

- https://github.com/personalrobotics/prpy/pull/343.
- https://github.com/personalrobotics/or_trajopt/pull/27

### DIFF
--- a/src/herbpy/herbrobot.py
+++ b/src/herbpy/herbrobot.py
@@ -33,7 +33,7 @@ class HERBRobot(Robot):
         from prpy.util import FindCatkinResource
 
         Robot.__init__(self, robot_name='herb')
-
+        self.robot_collision_checker = robot_collision_checker
 
         # Controller setup
         self.controller_manager = None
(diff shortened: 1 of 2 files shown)


src/herbpy/herbrobot.py
```diff
@@ -472,7 +478,6 @@ def SetStiffness(self, stiffness, manip=None):
     def Say(self, words, block=True):
         """Speak 'words' using talker action service or espeak locally in simulation"""
         if self.talker_simulated:
-            import subprocess
```
jeking04: Nit: Should we really move this to an import at the top? We can do a lot of things without `Say`.

src/herbpy/herbrobot.py
```diff
@@ -298,7 +305,6 @@ def __init__(self, left_arm_sim, right_arm_sim, right_ft_sim,
             self._say_action_client = SimpleActionClient('say', talker.msg.SayAction)
 
     def CloneBindings(self, parent):
-        from prpy import Cloned
```
jeking04: Nit: Should we really move this to the top. We can create a robot without needing clone functionality.

jeking04:
:+1:

PR #105, opened 2016-08-12
Changed post-processing algorithms to match ISER

mkoval:
This depends on https://github.com/personalrobotics/prpy/pull/347.


jeking04:
:+1: can confirm this matches the defaults we were setting in table clearing, which I believe are the ISER bests.

PR #106, opened 2016-08-12
Switched to ISER planning sequence

mkoval:
This depends on #105 and https://github.com/personalrobotics/prpy/pull/349.


jeking04:
This matches `table_clearing` branch `feature/log_for_benchmarks`:https://github.com/personalrobotics/table_clearing/blob/feature/log_for_benchmarks/src/table_clearing/robot_loader.py.

@gilwoolee can you confirm this is the final planner ordering we decided was best?

gilwoolee:
@jeking04 Yes, it is the final planner ordering. :D

PR #109, opened 2016-09-26
Re-implement MoveUntilTouch

ClintLiddick:
Requires https://github.com/personalrobotics/rewd_controllers/pull/5 and https://github.com/personalrobotics/pr_control_msgs/pull/1.

Fixes `wam.MoveUntilTouch` to work with the new controller, while following the original API. Tested working on HERB. Also includes a simple script to just run MoveUntilTouch directly after positioning end-effector by hand.

### DIFF
--- a/src/herbpy/wam.py
+++ b/src/herbpy/wam.py
@@ -314,7 +314,11 @@ class WAM(Manipulator):
 
             # execute movement
             kw_args['move_until_touch'] = True
-            traj = robot.ExecutePath(path, execute=execute, **kw_args)
+            with robot.CreateRobotStateSaver(
+                    Robot.SaveParameters.JointMaxVelocityAndAcceleration):
+                vl = robot.GetDOFVelocityLimits()
+                self.SetVelocityLimits(velocity_limit_scale*vl, 0.5)
+                traj = robot.ExecutePath(path, execute=execute, **kw_args)
             self.SetStiffness(True)
             return traj
 
--- a/src/herbpy/wam.py
+++ b/src/herbpy/wam.py
@@ -213,10 +213,11 @@ class WAM(Manipulator):
                        direction,
                        distance,
                        max_distance=1.0,
-                       max_force=5.0,
-                       max_torque=2.75,
+                       max_force=10.0,
+                       max_torque=0.6,
                        ignore_collisions=None,
                        velocity_limit_scale=0.25,
+                       execute=False,
                        **kw_args):
         """Execute a straight move-until-touch action.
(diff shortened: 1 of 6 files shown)


ClintLiddick:
This build is failing because of some damn reason @Shushman is looking into. @gilwoolee you can review and merge this PR asap, regardless of the build status.

src/herbpy/wam.py
```diff
@@ -236,54 +243,85 @@ def MoveUntilTouch(manipulator, direction, distance, max_distance=None,
...
+            if ns is '':
+                controller_name = 'move_until_touch_controller'
+            else:
+                controller_name = ns + '_move_until_touch_controller'
+
+            # load controller before setting threshold
+            robot.controller_manager.request([controller_name]).switch()
+
+            # connect to action server
+            server_name = '/' + controller_name + '/set_forcetorque_threshold'
```
gilwoolee: Is '..._move_until_touch_controller' always on by default?

src/herbpy/wam.py
```diff
@@ -223,10 +230,10 @@ def MoveUntilTouch(manipulator, direction, distance, max_distance=None,
         @param max_distance maximum distance in meters
         @param max_force maximum force in Newtons
         @param max_torque maximum torque in Newton-Meters
-        @param ignore_collisions collisions with these objects are ignored when 
+        @param ignore_collisions collisions with these objects are ignored when
```
gilwoolee: I think this name is misleading. Is this "continue colliding with this object(e.g. table) until max_force, max_torque are reached"? If so, MoveUntilTouch is dependent on actually colliding with this object, but the name seems to suggest that it's not relevant whether this object is touched or not. Unless my interpretation is mistaken, we should change this to something more like `collide_with`. If I am mistaken, this parameter needs clearer documentation.

src/herbpy/wam.py
```diff
@@ -223,10 +230,10 @@ def MoveUntilTouch(manipulator, direction, distance, max_distance=None,
...
         planning the path, e.g. the object you think you will touch
-        @param velocity_limit_scale A multiplier to use to scale velocity limits 
-        when executing MoveUntilTouch ( < 1 in most cases).           
+        @param velocity_limit_scale A multiplier to use to scale velocity limits
```
gilwoolee: Is this to slow down during MoveUntilTouch?

gilwoolee (commented):
This is great addition to herbpy! Thanks @ClintLiddick . 
I think most of the changes are straightforward. I have a few clarification questions.

PR #112, opened 2016-10-13
Use extend method for appending lists

ClintLiddick:


src/herbpy/action/blocks.py
```diff
@@ -40,7 +40,7 @@ def _GrabBlock(robot, blocks, table, manip=None, preshape=None,
         for b in blocks:
             # Get a TSR to move near the block.
             tsr_list = robot.tsrlibrary(b, 'grasp', manip=manip)
-            block_tsr_list += tsr_list
+            block_tsr_list.extend(tsr_list)
```
mkoval: What's the motivation behind this change? It looks uglier. :disappointed:

ClintLiddick: Eh, looking into it seems like the += is just as optimized as extend, and in this context both work. NBD then, I'll drop it.

PR #114, opened 2016-10-14
Generate documentation with Doxygen

brianhou:
This introduces a dependency on the `doxypy` module, which preprocesses the source code by moving the documentation from docstrings (which Doxygen doesn't understand) to Javadoc-style comments before the definitions.

I configured it so that Doxygen would use the repository README as the main page, but this required some style changes.
- Doxygen doesn't understand the GitHub Flavored Markdown syntax for syntax highlighting, e.g. <pre>```python</pre> This doesn't seem like a huge loss, since the feature wasn't consistently used and the only syntax highlighting was graying out comments.
- Relative links to source files on GitHub interfere with Doxygen's automatic linking. I've removed them for now, but it seems like a minor inconvenience when viewing the README on GitHub.


ClintLiddick:
Awesome! I'll merge this after I try it out

ClintLiddick:
Fixed to use the standard doxypy executable, and added it as a `doc_depend`. However, I don't think rosdep actually installs those, so it's just for show right now. Looks great otherwise! And even the markdown links work nicely!

PR #115, opened 2016-10-28
Remove rogue tests that are just testing motion planning

rachelholladay:
Most of the RoGuE tests are just running TSRs, which doesn't make a lot of sense because the point of their existence should not be to stress test our planners.


psigen:
Isn't the point of these tests to sanity check that the RoGuE function calls actually do something, and don't just crash due to silly stuff like committing syntax errors or changing the function arguments?

rachelholladay:
@psigen While I agree in theory, these tests end up actually testing if a TSR can be successfully planned. So theses tests are failing because planning failed, not because there is a syntax bug. One option would be to catch a planning failure, if you think that's a good idea?

psigen:
I guess it depends on why planning is failing.

* Is the TSR planning failing because it is non-deterministic?  If so, maybe it would be possible to plan without collision checking or similarly change the planning call to always succeed.
* If it is failing because of an upstream bug in TSRs, then failure is probably a legitimate result.

Otherwise, I suppose checking for a PlanningError might be alright, although then you risk missing bugs where an internal parameter is making one of your actions _always_ fail.

Shushman:
Merging to prevent herbpy build from failing. Test fixes can be seen to later.

PR #117, opened 2016-11-04
Rmh/demo poses

rachelholladay:
I should have PRed this when we made it. Its a demo script that used for switching between our HERB "picture poses". This is what we used at Nat'l Robotics Week.


ClintLiddick:
Awesome. Thanks!

PR #119, opened 2017-01-13
Feature/linting

rachelholladay:
Linted with pylint and formatted with yapf. The hope is that nothing was actually changed from a functionality standpoint, just better formatted. I tended to be more cautious in changes so we can go more aggressive if desired (ie I left unused function arguments because I wasn't sure if that would be a breaking change elsewhere). 

I also added the pylint configuration file. It was auto-generated with 3 messages to ignore:
-  C0103 : Invalid Name (because we have our own naming convention to match openrave)
- E1101 : Import has no member (because it doesn't interpret imports correctly)
- F0401 : Unable to import (because it says this even when it can import)
- E0611 : No name in module (see E1101)


psigen:
Hmm.. looking at the diff, I'm really not sure `yapf` helped the readability or maintainability of the code.

If the intent was to make automated style checking tests succeed, I think just doing this in an editor like SublimeText that has live pep8 and hand-formatting a few files at a time might produce a more usable result.

Can we do a first pass for `pylint` non-formatting-related warnings first and come back to the formatting in a second pass?

rachelholladay:
@psigen better?

scripts/inversedynamics_openrave.py
```diff
@@ -4,25 +4,25 @@
...
-        [ 5.147064052327087, -1.3622957134787639, -0.46951207447023835, 1.4621965585328196, -0.6525084419130015, 0.9042586427293137, -1.00526432576227 ],
-        [ 4.121337486391367, 0.5703433599319374, -2.2175569328355724, 2.049783564999927, -0.2955432490571725, -0.646366247533136, -1.4040613050409545 ],
-        [ 3.102831025390782, -0.04296665515610201, 0.1361467570187895, 0.045580105054784076, -3.0213214848284657, 0.0714728073293458, -0.24661498332101073 ]
+        [3.1670137672513565, -1.5896855774682224, -0.030490730254705722, 1.5286675450710463, -0.032386115821109854, -0.03605547869739324, -0.026511110707008654],
+        [5.147064052327087, -1.3622957134787639, -0.46951207447023835, 1.4621965585328196, -0.6525084419130015, 0.9042586427293137, -1.00526432576227],
+        [4.121337486391367, 0.5703433599319374, -2.2175569328355724, 2.049783564999927, -0.2955432490571725, -0.646366247533136, -1.4040613050409545],
+        [3.102831025390782, -0.04296665515610201, 0.1361467570187895, 0.045580105054784076, -3.0213214848284657, 0.0714728073293458, -0.24661498332101073]
```
psigen: Let's manually reformat this block specifically.

src/herbpy/action/blocks.py
```diff
@@ -62,9 +62,9 @@ def _GrabBlock(robot, blocks, table, manip=None, preshape=None,
             with env:
                 start_point = manip.GetEndEffectorTransform()[0:3, 3]
                 table_aabb = ComputeEnabledAABB(table)
-                table_height = table_aabb.pos()[2] + table_aabb.extents()[2]
+                #table_height = table_aabb.pos()[2] + table_aabb.extents()[2]
                 # 0.14 is the distance from finger tip to end-effector frame
-                current_finger_height = manip.GetEndEffectorTransform()[2,3] - 0.14
+                #current_finger_height = manip.GetEndEffectorTransform()[2, 3] - 0.14
```
psigen: Is this dead code?

src/herbpy/herbbase.py
```diff
@@ -29,19 +29,19 @@
...
 class HerbBase(MobileBase):
     def __init__(self, sim, robot):
         MobileBase.__init__(self, sim=sim, robot=robot)
         self.controller = robot.AttachController(name=robot.GetName(),
-#                                                 args='SegwayController {0:s}'.format('herbpy'),
-                                                 args='NavigationController {0:s} {1:s}'.format('herbpy', '/navcontroller'),
-                                                 dof_indices=[],
-                                                 affine_dofs=openravepy.DOFAffine.Transform,
-                                                 simulated=sim)
+#                             args='SegwayController {0:s}'.format('herbpy'),
```
psigen: Unused param?

src/herbpy/tsr/block.py
```diff
@@ -61,25 +61,25 @@ def block_at_pose(robot, block, position, manip=None):
...
+                 manip=manip_idx)
 
     place_tsr_chain = TSRChain(sample_start=False, 
-                               sample_goal = True,
-                               TSRs = [place_tsr, ee_tsr])
+                               sample_goal=True,
+                               TSRs=[place_tsr, ee_tsr])
     return [place_tsr_chain]
 
 @TSRFactory('herb', 'block', 'place_on')
```
psigen: Trailing whitespace here?

rachelholladay: Most of our files have a decent number of lines with trailing whitespace according to pylint. Is there an automated way you would recommend to take care of that? (I found some stuff online but was unsure what's best)

src/herbpy/action/blocks.py
rachelholladay: Well we never used this variable, but I was a bit conservative in that maybe people would want the reference of how to get the current finger height (I assume it was here for a reason at some point). 
I can just as easily go ahead and delete it.

src/herbpy/herbbase.py
rachelholladay: Yup, :+1:

src/herbpy/tsr/block.py
psigen: I think this is one of the better ones I've seen, you can give it a shot and check the diff (make sure you are all pushed on the off chance the `.git` files get messed up):
https://gist.githubusercontent.com/dpaluy/3690668/raw/585de899ed37a8ff95e0f2d4137e2e82e463a71c/git.sh

rachelholladay: It worked really well! Thanks! I pushed these updates

rachelholladay:
Fixed so that it matches what is updated in master. Are we good to go then?

PR #120, opened 2017-03-28
For usage after splitting out TSR from prpy

Shushman:
All the HERB tsr files have their imports corrected so that `rosrun herbpy console.py` now loads.
Please run other herbpy tests as well. The merge should be a 3-way merge of:

- `tsr` to include `remove_or_deps`
- `prpy` to include `split_tsr`
- `herbpy` to include this branch


brianhou (changes requested):
Tests are failing on Travis. You might need to specify the `tsr` module as a dependency in `package.xml`. Running them locally doesn't have that issue.

brianhou (approved):

